PR #63, opened 2016-05-05
Added self to editor list

cMadan:


arfon:
Thanks! Would you mind moving yourself up under Daniel S. Katz to keep things alphabetical?

cMadan:
Done. :)

PR #77, opened 2016-05-06
Once "then" is enough here.

Daniel-Mietchen:


arfon:
👍 agreed! Thanks for the edit.

PR #119, opened 2016-05-17
Update dependencies

xuanxu:
This PR updates codebase to use latest Rails & Octokit versions, and removes deprecation messages from RSpec.


arfon:
👍 thanks @xuanxu!

PR #161, opened 2016-08-20
Add atom link tag to layout

knewter:
This adds a link tag to the layout so that software can automatically
discover the RSS feed.  This is valuable when, for instance, pasting the
site's URL into a feed reader and letting the reader discover the feed's
existence.


arfon:
:+1: good idea. Thanks @knewter!

Issue #163, opened 2016-08-26
Encourage reviewers to open issues on submitted projects


jkahn:
I'd imagine you want to encourage technical back-and-forth with author and reviewer, but by doing it on the other project rather than the review thread, the reviewer's suggestions and improvements will appear in the changelog and issue tracker there and will thus be better acknowledged in the final publication edition.

Of course, mentioning those open issues in the review thread is helpful for documentation.

arfon:
Thanks for suggesting this, I agree this is what we want authors and reviewers to be doing.

Guidance on this probably makes sense to be added to the reviewer guidelines? What do you think @jkahn?

jkahn:
Sounds like a good place to add such a thing. Are the reviewer guidelines somewhere that I could submit a PR with suggested language change?

arfon:
> Sounds like a good place to add such a thing. Are the reviewer guidelines somewhere that I could submit a PR with suggested language change?

Sure thing! The reviewer guidelines are here: https://github.com/openjournals/joss/blob/master/app/views/home/about.html.erb#L128 and the template for the actual review issue is here (not sure where you think this is most appropriate): https://github.com/openjournals/joss/blob/master/app/views/shared/review_body.erb

Thanks in advance ⚡

jkahn:
I might file a separate pull request with changes to the review body, but this seems like a good start.

danielskatz:
I think this issue tries to solve a problem the wrong way.  If there is an issue that comes up in a review, I don't see a better place to discuss it than the review thread.  I think the concern that this issue was raised to address is traffic in the review thread that those other than the reviewer and author consider too much detail.  In my opinion, the better solution would be to limit those who see the review discussion by default.  I don't mind moving the discussion to a different thread within the review system, if that's a option, but moving it out of the review system and into the software being reviewed seems wrong to me, since this *is* an issue that came up in the review.

arfon:
> I think this issue tries to solve a problem the wrong way. If there is an issue that comes up in a review, I don't see a better place to discuss it than the review thread.

I'm sympathetic to both sides here. My concern though is that if the conversation is about something like installation help or technical discussion of this nature that discussion is of potential value to anyone who may use the software in the future. In my opinion that discussion is most discoverable on the actual software repository issue tracker rather than the JOSS review issue.

PR #165, opened 2016-08-29
Encourage reviewers to file issues, DOI after review

jkahn:
Encourage reviewers to file issues separately.

Also clarify that DOI should be set after review process is complete.

See https://github.com/openjournals/joss/issues/163#issuecomment-242853041


arfon:
👍 thanks @jkahn

PR #166, opened 2016-08-31
Update review body to discourage centithreads (#163)

jkahn:
Further work on https://github.com/openjournals/joss/issues/163

to try to avoid review-derailing or lengthy back-and-forth between reviewer and author.  (Rather, put that back-and-forth in its own issue and report only on its initiation and resolution within the review thread.)


arfon:
👍  thanks.

PR #190, opened 2016-10-24
Adds nirum as a reviewer

nirum:


arfon:
⚡️ many thanks @nirum

PR #236, opened 2017-03-20
Update papers_controller.rb

skrishnateja:
I think this is required for security. If not, just comment
How members are checked in routes?

### DIFF
--- a/app/controllers/papers_controller.rb
+++ b/app/controllers/papers_controller.rb
@@ -103,13 +103,17 @@ class PapersController < ApplicationController
 
   def start_meta_review
     @paper = Paper.find_by_sha(params[:id])
-
-    if @paper.start_meta_review!(nil, params[:editor])
-      flash[:notice] = "Review started"
-      redirect_to paper_path(@paper)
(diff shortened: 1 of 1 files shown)


arfon:
👋 @skrishnateja. Thanks for this PR and #237. I'm not sure they're necessary however.

- We already have a filter in `ApplicationController` that [checks for admins](https://github.com/openjournals/joss/blob/57533191e2bf7bd997ae38b5479eb36057bd1201/app/controllers/application_controller.rb#L13-L15)
- Which is then used in the `PapersController` [here](https://github.com/openjournals/joss/blob/57533191e2bf7bd997ae38b5479eb36057bd1201/app/controllers/papers_controller.rb#L4)

Perhaps I've misunderstood the intent here but I'm pretty sure this functionality is already covered.

pjotrp:
@skrishnateja you need to explain the purpose of your patches. Also take a look at the outstanding issues, maybe there is a quick win there.

arfon:
Yes, please take a look at the outstanding issues to see if there are some that you'd like to take on.

PR #237, opened 2017-03-20
Update user.rb isAdmin

skrishnateja:
This will make CI pass for the previous PR


arfon:
Closing because of my comment in https://github.com/openjournals/joss/pull/236#issuecomment-287767728

Issue #238, opened 2017-03-21
Consider adopting new terminology


csarven:
I'd like to propose:

* `s/submit/share`
* `s/submission/contribution`
* `s/paper/article`

Documented at http://csarven.ca/web-science-from-404-to-200#framing-our-vocabulary

PR #259, opened 2017-06-19
Commas

arfon:
Supersedes #238


danielskatz:
sorry, didn't find the template while searching the source tree, and having no knowledge of ruby on rails doesn't help...

arfon:
> sorry, didn't find the template while searching the source tree, and having no knowledge of ruby on rails doesn't help...

😁 No problem.

PR #268, opened 2017-07-13
Improve setting sha, don't break on collision

acrogenesis:
SecureRandom.hex can create duplicate values, this way you can avoid getting an error in such case.


arfon:
❤️ thanks for this @acrogenesis. Would you mind writing a small unit test to go along with this?

acrogenesis:
Sure, on it.

PR #273, opened 2017-07-14
Adding NumFOCUS links/messaging

arfon:
/ cc @christi3k for visibility.

This is how the [about](http://joss.theoj.org/about) page looks with this change:

screen shot 2017-07-14 at 9 27 38 am

...and the footer on every page now looks like this:

screen shot 2017-07-14 at 9 27 47 am


danielskatz:
should we say something on the about page -- about what this means?  e.g., if you want to donate to JOSS, you can do it via NUMFocus, and in the US, it's deducible as a charitable donation, etc...

tracykteal:
@danielskatz good idea to add a way to donate. NumFOCUS can get us a link we can put on the site for donations. It goes through FlipCause. Maybe we should have a 'Donate' page that specifically has this information, rather than putting it in About, since it's already rather long. It could be one of the top level menu items "Submit", "Papers", "About", "Donate".  Even though costs are low, there will continue to be a need for some operating costs, and it would be nice for people to be able to donate easily if they want to support the journal.

kyleniemeyer:
I agree that adding a new top-level page would be better than making "About" even longer.

kyleniemeyer:
Adding this new page shouldn't be hard, we just need the donation link to put there. Who do we need to ask about this?

tracykteal:
Leah or Lynn at NumFOCUS can get you the link.

kyleniemeyer:
Alright thanks, pursing this now—if I can get a donation link, then I'll add a page with this.

danielskatz:
@kyleniemeyer - any update on this?

kyleniemeyer:
No, I never heard anything back—will bug folks at NumFOCUS again.

arfon:
Merging this PR as-is. We can revisit donation stuff later.

PR #275, opened 2017-07-14
Lossless compression on images

acrogenesis:
Lossless compression on images using imageoptim. Saved 2.4MB

screen shot 2017-07-14 at 10 20 04 am


arfon:
😁 thanks!

PR #282, opened 2017-07-15
WIP: Business model -> Cost and Sustainability Model

### LINKED ISSUE #279: Suggest re-wording "Business Model" heading
"Business Model" generally refers to how you plan to make money out of the operation, so it doesn't quite apply to the information we provide there. Suggest changing the heading to *"Cost and Sustainability Model"* (and in fact adding a few words about sustainability).

arfon:
Eventually fixes #279 

/ cc @labarba and @kyleniemeyer


arfon:
@labarba, in #279 you suggested adding 'a few words about sustainability' to this section. If you suggest them here I'd be more than happy to update the PR with the changes 😁

PR #287, opened 2017-07-17
Add meta tags to header

kyleniemeyer:
Meta tags are needed for link previews on Slack, Twitter, Facebook, etc.


arfon:
👍 thanks for this @kyleniemeyer!

PR #313, opened 2017-08-16
Adding review question to existing author item: Is author list correct?

danielskatz:
adding question of if author list appears correct to existing author item in review


arfon:
👍 thanks @danielskatz

PR #321, opened 2017-08-28
Edit performance item

danielskatz:
changing text of the reviewer performance question to make it more clear that the reviewer can check the box if there are no performance claims


arfon:
Looks good. Thanks.

Issue #322, opened 2017-08-29
Update DOI URLs


arfon:
We currently use the old URL scheme for our DOI links (e.g. `http://dx.doi.org`) whereas we should be just using `https://doi.org` https://www.doi.org/doi_handbook/3_Resolution.html

kyleniemeyer:
This had been bugging me... thanks for making the issue.

arfon:
Between #324 and https://github.com/openjournals/whedon-api/commit/5ad482f659b1b2643c25c0f5a576eb72006262af this should now be fixed.

PR #324, opened 2017-09-01
[Solves #322] Update dx.doi.org to doi.org links 

arcticbarra:
Updated all links containing `http://dx.doi.org` to `https://doi.org`
https://github.com/openjournals/joss/issues/322


arfon:
✨ thanks @arcticbarra!

PR #355, opened 2017-11-13
Update Rails version

xuanxu:
This PR updates Rails to 4.2.10 version, the latest in the 4.2 series, an upgrade that includes several security issues and bug fixes.


arfon:
👍 thanks for this. The Rails update will be part of a larger set of changes coming up in #361 so I'm going to close this. Thanks for keeping us secure!

Issue #342, opened 2017-10-13
paper webpage should show date of publication


labarba:
When a visitor pulls up the page of an accepted paper, she cannot tell the date of publication. In fact, I don' t see how one could know the date of publication.

On the latest accepted paper:
http://dx.doi.org/10.21105/joss.00411
... if I click on the PDF, it has a date of Aug. 25. But if I click on the review issue, I see that it was accepted on Sep. 21.

rgieseke:
Is the date still in the current rendering of the PDF? Or is the comment above referring to the old layout?

arfon:
> Is the date still in the current rendering of the PDF? Or is the comment above referring to the old layout?

It's referring to the old layouts. We currently don't have any dates on the paper 😱

labarba:
Oops.

PR #363, opened 2017-12-05
Updated paper#show paper

### LINKED ISSUE #157: Citation Info
A few suggestions for paper pages, all in the spirit of making it as easy as possible for other researchers (who may not be computationally oriented) to cite the paper.
- Provide the citation in plain text on the main paper page. The doi2bib link is great for bibtex users, but this would add clarity and help non-bibtex people.
- The doi2bib link seems to pick up authors' middle initials as their first initials (at least [mine](http://www.doi2bib.org/#/doi/10.21105/joss.00036) does).
- Zotero's "Save to Zotero" plugin (Chrome, OS X) detects the citations in the paper as citations to grab, rather than identifying just the paper of interest. Not sure how hard it would be to fix this.

arfon:
This PR modifies the `paper#show` page on the JOSS site, e.g. what you see when you visit https://doi.org/10.21105/joss.00458

**This PR basically ditches the HTML render for the majority of the paper** and relies upon the native ability of modern browsers to display PDFs inline.

The motivations for making this change are numerous:

- The main HTML views for papers have long been an area of technical debt for JOSS. Currently the who page is served directly out of the database (yuck!) rather that being dynamically built. This means that these pages are _incredibly difficult to update_ and add new content e.g. https://github.com/openjournals/joss/issues/342 . With this change, updates to what these pages show will be much more straighforward.
- If authors include images in their papers, we have to upload them separately to GitHub and include a link to a web-hosted version of the figure. This is a hack and relies upon GitHub's goodwill to host file uploads. In the new views, the images are part of the PDF and as such we can remove this (potentially fragile) GitHub workaround.
- The current HTML views are IMO _ugly_. We did a bunch of work recently to improve the PDF layout of the papers. Developing a comprehensive set of CSS rules for rendering the HTML produced by Pandoc is a battle I'm not willing to commit to.

This isn't all 🦄 and 🌈 though. One major downside of this change is the mobile experience, i.e. if you visit the site on a mobile device, seeing an embedded PDF is less than ideal. I've added a [css rule](https://github.com/openjournals/joss/compare/pdf-js?expand=1#diff-1e07e367288f2ba9be554fce2ddceb8cR23) that basically hides the PDF embed on pages narrower than 600px wide.

**Before**
screen shot 2017-12-04 at 8 07 20 pm

**After**
screen shot 2017-12-05 at 9 57 07 am

I'm pretty convinced this is a solid improvement for JOSS but wanted to cc @openjournals/joss-editors for visibility.

Fixes https://github.com/openjournals/joss/issues/342
Fixes #157


labarba:
Great to have "Submitted" and "Accepted" dates right on there.
The only thing I don't like is the thin italic font of the paper title. 
Italics are rarely a good pick. Back to bold, I say.

Kevin-Mattheus-Moerman:
Looks great! :tada: I agree with @labarba about the bold font

labarba:
I just saw a page on my phone, via a tweet link, and think the embedded PDF looks fine.

labarba:
image1

arfon:
> Would it be much work to add something like (e.g. like Zenodo has it):

@Kevin-Mattheus-Moerman - I'd like to add something like that but I didn't get a chance to add it to this set of changes.

PR #364, opened 2017-12-06
Upgrade to Rails 5.1

parrish:
This upgrades Rails and the gemfile in general.  AASM and Octokit seem to have introduce breaking changes on a minor release, so I left those as-is for the moment.


arfon:
@parrish this is now deployed to http://staging.joss.theoj.org

PR #369, opened 2017-12-11
[WIP] Add paper searching

parrish:
This PR sets up a postgres-based full text search with "faceting" for the paper status (it's basically just piggy-backing on the existing index view with an optional search term).  I'm using a postgres trigger to handle updating the cached lexemes in a tsvector column on the paper model.

A few caveats

- ~0 UX.  I basically just stuck a box on the page and you can hit enter to search.~
- Using a trigger for this is great, but it requires a switch to structure.sql instead of schema.rb
  - Heroku _might_ take some coercion to get this deploying seamlessly


arfon:
Closing this in favour of #575

Issue #176, opened 2016-09-26
Handling multiple reviewers


Kevin-Mattheus-Moerman:
Our current review pipeline handles a single reviewer. We need to upgrade the process to handle N editor assigned reviewers. Perhaps this requires an adjustment of the current tick box style form. Are multiple tick boxes required, one for each reviewer per point, or do we use multiple forms, one for each reviewer? What does the reviewer system look like for multiple reviewers? 

For multiple reviewers, should we collect reviewer feedback as editors first and then present it to the authors? E.g. what if reviewer 1 comments "_Please implement A_", followed by reviewer 2 who comments "_Whatever you do please never implement A_". If reviewer 1's comment comes first and reviewer 2's comment several days later, perhaps the authors will have started implementing A already. So do we need to formalize the way the editor recommends how the authors address the reviewers comments?

labarba:
I would continue the review conversations openly on the Issue thread, rather than separately with the editor. If reviewers disagree on something, that is part of the conversation. The editor might make a decision to favor one of the reviewers on the issue of disagreement. But doing it all publicly has value.

kyleniemeyer:
I want to bump this issue, as I think it's quite important to support >1 reviewer—I think in most cases, we should want more than one person to review. It shouldn't be too hard to have a list of names, rather than just one.

labarba:
When adding a second reviewer, is it possible to add a second checkbox to each item in the Review Checklist?

danielskatz:
or a second checklist for reviewer2 would also work and would be more likely to be possible in GItHub

iljah:
A separate issue in joss-reviews would a good way to start at least I think. This way both reviewers can do their work without accidentally messing with wrong check boxes, etc. Perhaps rename the title of issue from [REVIEW] to [REVIEW1], [REVIEW2], etc.

kyleniemeyer:
@iljah One of the strong features of JOSS we don't want to lose is the single review issue, where the review is more of a conversation between the reviewer(s), editor, and author. Moving to two review issues would simplify the check box question, but lose that combined discussion.

@arfon @parrish any updates on handling more than one reviewer with whedon?

arfon:
> @arfon @parrish any updates on handling more than one reviewer with whedon?

:Wave: @parrish - would you be able to take a look at this. Here's the code for handling the reviewer assignments: https://github.com/openjournals/whedon-api/blob/master/whedon.rb#L223-L249

Basically I think we need Whedon to be able to:

[...]

PR #389, opened 2018-03-17
[WIP]: Adding ability to handle multiple reviewers

arfon:
/ cc https://github.com/openjournals/joss/issues/176


app/controllers/papers_controller.rb
```diff
@@ -91,7 +91,7 @@ def start_review
   def api_start_review
     if params[:secret] == ENV['WHEDON_SECRET']
       @paper = Paper.find_by_meta_review_issue_id(params[:id])
-      if @paper.start_review!(nil, params[:editor], params[:reviewer])
+      if @paper.start_review!(nil, params[:editor], params[:reviewers])
```
arfon: `reviewers` is now a csv string, i.e. for one review would be `reviewer1` and for two reviewers would be `reviewer1,reviewer2`

PR #390, opened 2018-03-18
Add bootsnap

parrish:
I was feeling impatient with specs, so I added [bootsnap](https://github.com/Shopify/bootsnap) for snappier boots.  I also noticed the spring binstubs weren't installed, but I wasn't sure if you were opposed -- I'll add them here if you're into it.

Running `$ time rspec 'spec/models/paper_spec.rb[1:1]'`

Before: `real    0m4.723s`

After: `real    0m2.079s`

With spring: `real    0m0.747s`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,7 @@
 source 'https://rubygems.org'
 
 gem 'aasm', '~> 4.1.0'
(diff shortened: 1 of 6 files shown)


arfon:
👍 LGTM

Issue #219, opened 2017-01-12
What's our policy on issue-trackers that require registration and login?


labarba:
I'm seeking comment from the JOSS Editors: In this submission, the project owners have indicated that they will accept issues, not through the GitHub issue tracker, but through their separate platform, requiring registration and login. This bothers me ... but I want to know what others think:
https://github.com/openjournals/joss-reviews/issues/151#issuecomment-272135155

arfon:
> This bothers me ... but I want to know what others think

I agree that this is less than ideal, however I'm not sure it's reasonable for us to mandate where people do their issue tracking (or code-hosting for that matter). Perhaps in this case the submitting authors could register some kind of 'guest' account for the reviewer to use?

labarba:
What bothers me is that they request *registration and login* to submit an issue.

kyleniemeyer:
I also agree that using GitHub (or a similar commonly used platform) would be better than something custom. However, such services also require registration and login to submit issues—it's just that people more commonly have accounts at these places. (Also, I suppose, users trust their personal information with the service rather than the specific project owners.)

I think this is more something that could come out of the review process as suggestions for improving ways to contribute and submit issues, rather than being something that would make their submission ineligible for publication in JOSS. I agree with @arfon that some sort of guest account would make submitting issues easier—you don't want to make it hard to do this.

(All that said, if this is something we care about, we should add it to the submission guidelines. Personally I can't imagine moving a project off GitHub, BitBucket, or GitLab to something custom because that sounds like a nightmare, but there could conceivably be some institutional reasons for it.)

labarba:
Good points, @kyleniemeyer !

For some reason, I feel differently about having to log in to GitHub—the hosting service—than giving the software project my details. It just does not feel as "open" ...

kyleniemeyer:
@labarba I totally agree with you on that!

jflorence:
I agree that submitting an issue to an open source project should be as simple and fast as possible, in order not to deter people from doing so, and improve the quality of the software as much as possible.
Requiring a login might be necessary to avoid spam and other kinds of problems. 
But then it would be better to host the issue tracker on a website like github, where bug reporters might be very likely to already have an account.

iljah:
> What bothers me is that they request registration and login to submit an issue.

[...]

PR #397, opened 2018-04-18
Adding requirements for repositories

arfon:
As per the JOSS editors discussion, here are the changes for #219

/ cc @openjournals/joss-editors


danielskatz:
change "you" to "your" in line 10 please

arfon:
Good catch. Thanks.

PR #404, opened 2018-05-03
Add message about review "deadline"

### LINKED ISSUE #312: Make a reviewer deadline
Sorry if this isn't the right place to make a "process" issue, but it felt right...

Perhaps JOSS should have a reviewer deadline?  Many of the academic journals I'm used to set a clock (2 weeks, 4 weeks, etc.) from when the reviewer accepts.  Of course we all know these aren't "real" deadlines in that usually they are flexible... but having the deadline has a psychological impact, at least for some reviewers, that makes them do it in a timely (or at least predictable) manner.

marwahaha:
fixes #312 

This was @jmjwozniak's suggestion to resolve @eteq 's process issue.

Sorry if I revived an old debate ..


danielskatz:
While I appreciate your trying to address this idea, I don't think we actually should do this, and instead, I think we should just close #312 and say that we don't plan to do anything currently.

PR #405, opened 2018-05-03
Pad paper details with lists

### LINKED ISSUE #292: Website: bullet list in submission "Description" breaks alignment
screen shot 2017-07-20 at 10 06 52 am

As you can see, when the submitter includes a bullet list in their "Description," it breaks the left-alignment of the content on the journal's web page. 

This is a small thing, but it looks off.

marwahaha:
fixes #292 

I couldn't get the DB running, so I couldn't actually test this locally. I made this change in-browser and they seemed to get this behavior:
image

### DIFF
--- a/app/assets/stylesheets/home.scss
+++ b/app/assets/stylesheets/home.scss
@@ -324,6 +324,12 @@
   .list-paper > .details > .paper-title {
     font-size: 16px;
   }
+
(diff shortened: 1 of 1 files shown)


arfon:
Thanks for working on this @marwahaha 🙇. With the updated web UI this update isn't really necessary any longer.

PR #408, opened 2018-05-17
Add Jed to editorial board

jedbrown:
I noticed that the Rails upgrade changed the format relative to #309, but I think this is correct.


arfon:
👋@jedbrown. Sorry for being slow to respond here. I'll work on this later today or tomorrow.

PR #421, opened 2018-07-01
text smoothing

danielskatz:


arfon:
You have the best words @danielskatz 😄

PR #426, opened 2018-07-15
Reviewer invite letter: Use updated link to doc site

jedbrown:


jedbrown:
Alternative is to link https://joss.readthedocs.io/en/latest/reviewer_guidelines.html but it is too vague and requires another click to see criteria. Of course deep-linking to the criteria means reviewers will rarely read the philosophy page.

PR #438, opened 2018-08-18
Basic metrics dashboard

arfon:
This PR adds a collection of basic metrics views show high level stats for papers going through the JOSS system. 

screen shot 2018-08-18 at 7 45 27 am
screen shot 2018-08-18 at 7 46 07 am
screen shot 2018-08-18 at 7 46 17 am

### DIFF
--- a/app/models/paper.rb
+++ b/app/models/paper.rb
@@ -6,6 +6,8 @@ class Paper < ActiveRecord::Base
               :validate => true,
               :foreign_key => "user_id"
 
+  belongs_to  :editor
+
   include AASM
 
   aasm :column => :state do
@@ -53,6 +55,8 @@ class Paper < ActiveRecord::Base
   default_scope  { order(:created_at => :desc) }
   scope :recent, lambda { where('created_at > ?', 1.week.ago) }
   scope :submitted, lambda { where('state = ?', 'submitted') }
+
+  scope :since, -> (date) { where('accepted_at >= ?', date) }
   scope :in_progress, -> { where(:state => IN_PROGRESS_STATES) }
   scope :visible, -> { where(:state => VISIBLE_STATES) }
   scope :everything, lambda { where('state NOT IN (?)', ['rejected', 'withdrawn']) }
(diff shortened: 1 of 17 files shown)


app/models/paper.rb
```diff
@@ -23,11 +25,11 @@ class Paper < ActiveRecord::Base
     end
 
     event :start_meta_review do
-      transitions :from => :submitted, :to => :review_pending, :after => :create_meta_review_issue
+      transitions :from => :submitted, :to => :review_pending, :if => :create_meta_review_issue
```
arfon: This conditional makes sure that the AASM transition only happens _if_ the `create_meta_review_issue` method returns `true`

app/models/review_issue.rb
```diff
@@ -86,8 +86,8 @@ def editor
     body.match(/\*\*Editor:\*\*\s*(@\S*|Pending)/i)[1]
   end
 
-  # Extract the Reviewer from the issue body
-  def reviewer
-    body.match(/\*\*Reviewer:\*\*\s*(@\S*|Pending)/i)[1]
+  # Extract the Reviewers from the issue body
+  def reviewers
```
arfon: This just matches the logic in [WhedonAPI](https://github.com/openjournals/whedon-api/blob/b0720d8c30673b87c5bb8860d361b403338b31e9/whedon.rb#L266-L268)

spec/controllers/papers_controller_spec.rb
```diff
@@ -57,26 +77,53 @@
...
+      editor = create(:editor, :login => "mouse")
+      editing_user = create(:user, :editor => editor)
+
+      paper = create(:review_pending_paper, :state => "review_pending", :meta_review_issue_id => 1234, :user_id => user.id)
+      fake_issue = Object.new
+      allow(fake_issue).to receive(:number).and_return(1)
+      allow(GITHUB).to receive(:create_issue).and_return(fake_issue)
+
+      # Can't create review issue because editor is invalid
+      expect {
```
arfon: Actually testing these transitions now...

lib/tasks/utils.rake
```diff
@@ -0,0 +1,51 @@
+require "util/console_extensions"
```
arfon: There are a couple of data migrations here.

leeper:
Looks great!

arfon:
OK this is now live at https://joss.theoj.org/dashboard . You can only see this if you're an editor.

PR #448, opened 2018-09-06
LINK: remove trailing slash from JOSS DIO link

seberg:
At least when I clicked in on my browser the doi website did not work. Remove the slash forwards to the correct page.


arfon:
👍 thanks @seberg

Issue #434, opened 2018-08-11
Display ORCiD logo/links?


leeper:
Given we capture ORCiD for authors, should we add ORCiD links on papers and on paper HTML pages? Seems lots of other journals are doing this.

arfon:
These are now visible on the `paper#show` pages, e.g. https://joss.theoj.org/papers/10.21105/joss.01601 :

<img width="1204" alt="Screen_Shot_2019-08-12_at_2_32_19_PM" src="https://user-images.githubusercontent.com/4483/62888911-4d903900-bd0e-11e9-9bf9-1143509e5eda.png">

PR #458, opened 2018-10-19
[WIP] Add authors information to accepted papers pages

bryanwweber:
This PR is WIP to add the ORCID/author names as suggested in #434. @arfon suggested there be a new column in the `papers` table to store the authors from the Markdown paper header. I'm not sure where the `paper.md` is read, or how the `params` hash is filled with, e.g., the `citation_string`. There also already appears to be an `authors` key in this hash, e.g., https://github.com/openjournals/joss/blob/30f70cd53961521da272ab6d28e027c3c201c1c0/app/controllers/papers_controller.rb#L115 So, my question is, how do I get the data to serialize it, and how do I put that into the database? Thanks 😄

### DIFF
--- a/db/migrate/20181019161622_add_authors_column.rb
+++ b/db/migrate/20181019161622_add_authors_column.rb
@@ -1,5 +1,5 @@
 class AddAuthorsColumn < ActiveRecord::Migration[5.1]
   def change
-    add_column :papers, :authors:, :text
+    add_column :papers, :authors, :text
   end
 end
--- a/db/migrate/20181019161622_add_authors_column.rb
+++ b/db/migrate/20181019161622_add_authors_column.rb
@@ -1,0 +1,5 @@
+class AddAuthorsColumn < ActiveRecord::Migration[5.1]
+  def change
+    add_column :papers, :authors:, :text
+  end
+end


db/migrate/20181019161622_add_authors_column.rb
```diff
@@ -0,0 +1,5 @@
+class AddAuthorsColumn < ActiveRecord::Migration[5.1]
+  def change
+    add_column :papers, :authors:, :text
```
arfon: Looks like there's a small typo here:

```suggestion
    add_column :papers, :authors, :text
```

arfon:
Thanks for starting this @bryanwweber. 

>  I'm not sure where the `paper.md` is read or how the params hash is filled with, e.g., the `citation_string`

This is actually generated by the `Whedon` command line Rubygem that we use in our editorial tooling: https://github.com/openjournals/whedon/blob/master/lib/whedon/processor.rb#L165-L185

When `Whedon` builds the metadata to deposit with JOSS, the environment has access to the `paper.md` file so this would be a good moment to grab the `authors` from `paper.md` and serialize it to be passed back to JOSS.

A couple of wrinkles, it looks like converting a Hash (dictionary in Python) is a bit funky (e.g. https://stackoverflow.com/questions/798710/ruby-how-to-turn-a-hash-into-http-parameters) but possible.

As a suggested next step, what do you think about investigating how to URL encode the authors YAML someway? As a reminder, the `paper.md` YAML looks like [this example paper](https://joss.readthedocs.io/en/latest/submitting.html#example-paper-and-bibliography):

```yaml
authors:
  - name: Adrian M. Price-Whelan
    orcid: 0000-0003-0872-7098
    affiliation: "1, 2" # (Multiple affiliations must be quoted)
  - name: Author 2
    orcid: 0000-0000-0000-0000
    affiliation: 2
```

I *think* all we want to do here, is parse this YAML [here's an example of how to open a YAML file](https://github.com/openjournals/whedon/blob/master/lib/whedon.rb#L68) and create key/value pairs that we URL encode, i.e. the above should turn into:

```
{ 
  "Adrian M. Price-Whelan" => "0000-0003-0872-7098", 
  "Author 2" => "0000-0000-0000-0000" 
}
```

Once we've figured out this step, we can modify the `Whedon` gem to do this processing and then the logic on the JOSS application to handle the passed authors/ORCIDs. Does that sound OK?

> https://github.com/openjournals/joss/blob/30f70cd53961521da272ab6d28e027c3c201c1c0/app/controllers/papers_controller.rb#L115

Minor comment: I don't think we currently use the `authors` param here so we can probably leave this incoming API method as-is and pass the author ORCIDs etc. as this named parameter once we have figured it out.

bryanwweber:
@arfon I think that makes sense. 2 thoughts

1. Should we do the modifications in Whedon first, to figure out what the URL encode will look like?
2. Should we also try to encode the affiliations, so we can print the authors list in the HTML with their affiliations as well? That obviously gets quite a bit more complicated, so maybe we can just get the ORCID working first, and then add the affiliation

arfon:
> 1. Should we do the modifications in Whedon first, to figure out what the URL encode will look like?

Sure thing. That's not a bad idea. Are you OK to clone that project and see how far you get?

I would start by looking at the code that executes when someone types: 
```
bundle exec whedon deposit 1234 <- where this last number is the review issue number
```

This code is here: https://github.com/openjournals/whedon/blob/master/bin/whedon#L89-L109 and the `deposit` method to JOSS is here: https://github.com/openjournals/whedon/blob/master/lib/whedon/processor.rb#L165-L185 . The `paper` object here should already know about its authors, and they are already abstracted as a class [here](https://github.com/openjournals/whedon/blob/master/lib/whedon/author.rb). So we can probably do something like:

```ruby
authors = []
paper.authors.each do |a|
  authors << { 'name' => a.name, 'orcid' => a.orcid, 'affiliation' => a.affiliation } 
end
```



> 2. Should we also try to encode the affiliations, so we can print the authors list in the HTML with their affiliations as well? That obviously gets quite a bit more complicated, so maybe we can just get the ORCID working first, and then add the affiliation

Possibly. Although I'm wondering if we should potentially base64 encode the string/Hash (or something similar), that way, we don't have to worry about crazy URL encoding of a nested structure.

bryanwweber:
> Are you OK to clone that project and see how far you get?

Yup! I'll make a PR over there once I get started.

>  potentially base64 encode the string/Has

Would the standard library's [`encode_www_form`](https://ruby-doc.org/stdlib-2.1.0/libdoc/uri/rdoc/URI.html#method-c-encode_www_form) work? Could hypothetically replace the "1, 2" with the actual affiliation names before sending. That'd be a long URL though...

arfonsmith:
Ok great! On second thoughts, let’s skip the affiliation information for
now. I think the author name and ORCID are the important things to capture.

Plus, this way we can just pass key/value pairs of author name => orcid

arfon:
:wave: @bryanwweber, as I mentioned over in https://github.com/openjournals/whedon/pull/40#issuecomment-518050875, there's now a `metadata` field available on `Paper` which has the author information present as well as the ORCIDs etc.

If you're still interested in working on this PR then it might be a good idea to start in a fresh PR as there's been so many changes to the site since this PR was opened. Also, if you don't have the bandwidth to work on this that's OK too :-)

bryanwweber:
Thanks @arfon, sorry I couldn't get to this. Glad it's been added now!

arfon:
> Thanks @arfon, sorry I couldn't get to this. Glad it's been added now!

No problem at all. I'm sorry I was able to make myself more available to help you towards making this contribution 😞.

PR #460, opened 2018-11-01
updating my info

danielskatz:
It's not clear to me that seeds.db is an intuitively named place to store editor information.


arfon:
> It's not clear to me that seeds.db is an intuitively named place to store editor information.

Agreed, its not. There's actually an admin UI where I can update this information so we don't need to change the data here. 

`seeds.db` is a Rails convention for a place to put data when you want to first initialize an application.

PR #481, opened 2019-01-04
Link from submission workflow to review criteria

arokem:
Maybe this extra nudge would help?


arokem:
To clarify: as an author, I would not expect to have to look at the "reviewer documentation". There is already a link to this at the bottom, but it's framed as something optional.

arfon:
> Maybe this extra nudge would help?

Can't hurt...

PR #494, opened 2019-02-06
Mention open access

### LINKED ISSUE #493: about page doesn't say "open access"?
It seems like we should highlight the open aspects of the journal itself on http://joss.theoj.org/about#about, including the fact that papers are open access - this is kind of buried at the bottom under content licensing now.

arfon:
Fixes #493 

/ cc @danielskatz


danielskatz:
thanks

PR #508, opened 2019-03-05
adding contact info to JOSS about page

danielskatz:


danielskatz:
👋 @arfon - how long should it take before this appears in the live page?

arfon:
It's now live.

PR #513, opened 2019-03-17
update my bio

lheagy:
Thanks for pointing this out @labarba!


labarba:
Done the edit via the website GUI.

PR #520, opened 2019-04-08
docs/submitting: remove reference to old Zenodo version

jedbrown:
Upon acceptance, authors archive their repositories to Zenodo, but
that will be a new version not available when writing the paper.  In
the provided example, the `@zenodo` reference in the paper is for v0.2.1
while the archive in the sidebar is v0.2.2 (the post-review version).
The citation to the old version does not add value, but may cause
confusion.

Reported-by: Gregory R. Lee <grlee77@gmail.com>


arfon:
Make sense. Thanks for spotting this.

PR #524, opened 2019-04-10
Specify that links to issues should be in comments

leouieda:
Just had a reviewer edit the issue body to include links to issues they raised on the target repository.
It seems they interpreted "review thread" as the checklist (which is right below those instructions). This PR clarifies that the links should be posted as comments on the review thread to avoid future misunderstandings.

### DIFF
--- a/app/views/shared/review_body.erb
+++ b/app/views/shared/review_body.erb
@@ -18,7 +18,7 @@ Markdown: [![status](<%= url %>/papers/<%= paper.sha %>/status.svg)](<%= url %>/
 ```
 **Reviewers and authors**:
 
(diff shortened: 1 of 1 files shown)


arfon:
:+1: thanks @leouieda

jedbrown:
FWIW, it's typically (for repositories on GitHub) easier and cleaner to mention the review thread (paste the URL) when creating issues, in which case GitHub will automatically notify the review thread in a way that tracks its status (like when the issue or PR is closed/merged).

leouieda:
@jedbrown I agree and that's usually what I ask reviewers to do. But I like having them post a comment linking to the issue in the review thread as well so that we all get notifications and act as soon as possible.

jedbrown:
Fair enough. The author of the package presumably gets notified when the issues/PRs are filed on their repository, but if you would like the editor and other reviewers to also be notified, then I suppose posting in the review thread makes sense. (I'm not sure that's necessary. I've had reviewers disable notification from a review issue due to notification overload.)  It's a little complicated to explain the request for both.

PR #528, opened 2019-04-23
review_body: link to new site (joss.readthedocs.io)

jedbrown:
@arfon I recall discussing running joss.readthedocs.io on our custom domain, but in lieu of implementing that, our issues should point to where the instructions actually reside.


arfon:
👍 thanks @jedbrown

PR #538, opened 2019-05-31
about page cost model update

danielskatz:
adding website domain cost
changing calculation to 300 papers per year


arfon:
👍 thanks.

PR #539, opened 2019-05-31
Add example citation commands

kyleniemeyer:
There are multiple citation commands that change the in-text format, and we should document these in an obvious place.


danielskatz:
+1

Issue #442, opened 2018-08-21
Why is "version" a field in the submission?


arokem:
Version information often changes during the review (e.g., when proposed changes are integrated into the software, and a new version is made as part of the review process). Besides, the version information doesn't make it to the final paper. Why are we even asking for it? 

See also: https://github.com/maxrousseau/pfla/issues/2

labarba:
I agree this is awkward! It's even a required field when submitting; at least make it optional.

jedbrown:
I think the version is counterproductive at submission time (allowing a
non-default branch would be nice), but that the post-review version to
be archived should probably be (preferably signed) tag.  Could we teach
whedon to perform consistency checks/automatically identify the version
from the DOI?

danielskatz:
I'm going to close this, in favor of https://github.com/openjournals/joss/issues/857 where a related discussion happened.

PR #542, opened 2019-06-11
Proposed changes to checklist

Kevin-Mattheus-Moerman:
I did the following: 

1. Added additional requirements for paper. 
2. Removed version tick box.
3. Updated the text about authorship and contributions. I attempted to use the following to link to the project contributors: 
```Ruby
<%= link_to “contributors”, paper.repository_url + “/graphs/contributors”, :target => “_blank” %>
```
I also merged the authorship point from the paper section with this point. 
4. I removed the question about authors/affiliations. I propose we let Whedon check for affiliations instead, perhaps based on ORCID, and to list missing affiliations if any. It feels odd to ask the reviewers to check the affiliations, i.e. they'd need to google the authors and check where they work. Not sure if that is their job. 

I feel perhaps the References point (asking about DOI's) can be removed too since Whedon checks that at the end. But perhaps we want to start the DOI discussion earlier on in the review process so I left it for now.

### DIFF
--- a/app/views/content/github/_review_checklist.erb
+++ b/app/views/content/github/_review_checklist.erb
@@ -10,8 +10,7 @@
 
 - [ ] **Repository:** Is the source code for this software available at the <%= link_to "repository url", paper.repository_url, :target => "_blank" %>?
 - [ ] **License:** Does the repository contain a plain-text LICENSE file with the contents of an [OSI approved](https://opensource.org/licenses/alphabetical) software license?
-- [ ] **Version:** Does the release version given match the GitHub release (<%= paper.software_version %>)?
-- [ ] **Authorship:** Has the submitting author (<%= paper.submitting_author.github_username %>) made major contributions to the software? Does the full list of paper authors seem appropriate and complete?
+- [ ] **Contribution and authorship:** Based on the list of <%= link_to "contributors", paper.repository_url + "/graphs/contributors", :target => "_blank" %> to this project, can you confirm: 1) that the submitting author (<%= paper.submitting_author.github_username %>) made major contributions to the software?, and 2) That the full list of paper authors seems appropriate and complete?
 
 ### Functionality
 
@@ -25,11 +24,13 @@
 - [ ] **Installation instructions:** Is there a clearly-stated list of dependencies? Ideally these should be handled with an automated package management solution.
 - [ ] **Example usage:** Do the authors include examples of how to use the software (ideally to solve real-world analysis problems).
 - [ ] **Functionality documentation:** Is the core functionality of the software documented to a satisfactory level (e.g., API method documentation)?
-- [ ] **Automated tests:** Are there automated tests or manual steps described so that the function of the software can be verified?
+- [ ] **Automated tests:** Are there automated tests or manual steps described so that the functionality of the software can be verified?
 - [ ] **Community guidelines:** Are there clear guidelines for third parties wishing to 1) Contribute to the software 2) Report issues or problems with the software 3) Seek support
 
 ### Software paper
 
-- [ ] **Authors:** Does the `paper.md` file include a list of authors with their affiliations?
+- [ ] **Summary:** Has a clear description of the high-level functionality and purpose of the software for a diverse, non-specialist audience been provided?
 - [ ] **A statement of need:** Do the authors clearly state what problems the software is designed to solve and who the target audience is?
+- [ ] **State of the art:** Are key relevant or alternative approaches cited and briefly discussed?
+- [ ] **Quality of writing:** Is the paper well written (i.e. such that it does not require editing for structure, language, and writing quality)?
 - [ ] **References:** Do all archival references that should have a DOI list one (e.g., papers, datasets, software)?


app/views/content/github/_review_checklist.erb
```diff
@@ -10,8 +10,7 @@
 
 - [ ] **Repository:** Is the source code for this software available at the <%= link_to "repository url", paper.repository_url, :target => "_blank" %>?
 - [ ] **License:** Does the repository contain a plain-text LICENSE file with the contents of an [OSI approved](https://opensource.org/licenses/alphabetical) software license?
-- [ ] **Version:** Does the release version given match the GitHub release (<%= paper.software_version %>)?
-- [ ] **Authorship:** Has the submitting author (<%= paper.submitting_author.github_username %>) made major contributions to the software? Does the full list of paper authors seem appropriate and complete?
+- [ ] **Contribution and authorship:** Based on the list of <%= link_to "contributors", paper.repository_url + "/graphs/contributors", :target => "_blank" %> to this project, can you confirm: 1) that the submitting author (<%= paper.submitting_author.github_username %>) made major contributions to the software?, and 2) That the full list of paper authors seems appropriate and complete?
```
arfon: So my only problem with this is that the link will only work if the source code is on GitHub. We _prefer_ submissions that are on GitHub and stored in git but adding URLs like this makes this even more locked in. Thoughts?

app/views/content/github/_review_checklist.erb
```diff
@@ -25,11 +24,13 @@
 - [ ] **Installation instructions:** Is there a clearly-stated list of dependencies? Ideally these should be handled with an automated package management solution.
 - [ ] **Example usage:** Do the authors include examples of how to use the software (ideally to solve real-world analysis problems).
 - [ ] **Functionality documentation:** Is the core functionality of the software documented to a satisfactory level (e.g., API method documentation)?
-- [ ] **Automated tests:** Are there automated tests or manual steps described so that the function of the software can be verified?
+- [ ] **Automated tests:** Are there automated tests or manual steps described so that the functionality of the software can be verified?
 - [ ] **Community guidelines:** Are there clear guidelines for third parties wishing to 1) Contribute to the software 2) Report issues or problems with the software 3) Seek support
 
 ### Software paper
 
-- [ ] **Authors:** Does the `paper.md` file include a list of authors with their affiliations?
```
arfon: We can't guarantee that we will have access to affiliation information using ORCID as this very much depends on whether the user has added this information to their profile _and_ whether the allow this information to be read publicly via the API (the default I think is to *not* allow this). 

So, I'm happy to remove this if we think that the `Contribution and authorship` question covers the authorship question.

app/views/content/github/_review_checklist.erb
```diff
@@ -25,11 +24,13 @@
...
+- [ ] **Summary:** Has a clear description of the high-level functionality and purpose of the software for a diverse, non-specialist audience been provided?
 - [ ] **A statement of need:** Do the authors clearly state what problems the software is designed to solve and who the target audience is?
+- [ ] **State of the art:** Are key relevant or alternative approaches cited and briefly discussed?
```
arfon: I'm not sure `State of the art` reflects our [submission guidelines](https://joss.readthedocs.io/en/latest/submitting.html#submission-requirements). What about `**Significant contribution to open source: ** Is this software a significant contribution to the available open source software in this research area?`

app/views/content/github/_review_checklist.erb
Kevin-Mattheus-Moerman: I see the problem. Did not think about that. We can remove the URL so.

app/views/content/github/_review_checklist.erb
Kevin-Mattheus-Moerman: I think so. I also feel this is perhaps a point the editors should check rather than the reviewers.

app/views/content/github/_review_checklist.erb
Kevin-Mattheus-Moerman: Actually you are raising a different point. What I meant by this "State of the art" point is whether or not the authors describe the current state of the art in their paper. Currently our guidelines say there should be: "A list of key references including a link to the software archive". First of all we no longer should ask for a link to the software archive (will be added automatically), second of all we do not just want a "list of key references" we want them to write a nice paragraph where they summarize what I called the "state of the art", or more specifically "Are key relevant or alternative approaches cited and briefly discussed". So this statement was aimed at replacing the references statement and aims to hint more at the quality of writing. 
Your point on significance of the work in general can also be added but perhaps that is captured by the statement of need stuff anyway. Thoughts?

Kevin-Mattheus-Moerman: Any thoughts on this. Let me know if I can help or if we should chat about this.

arfon:
@danielskatz @kyleniemeyer @labarba - could you take a look at these changes and see what you think?

app/views/content/github/_review_checklist.erb
```diff
@@ -25,11 +24,13 @@
...
+- [ ] **Automated tests:** Are there automated tests or manual steps described so that the functionality of the software can be verified?
 - [ ] **Community guidelines:** Are there clear guidelines for third parties wishing to 1) Contribute to the software 2) Report issues or problems with the software 3) Seek support
 
 ### Software paper
 
-- [ ] **Authors:** Does the `paper.md` file include a list of authors with their affiliations?
+- [ ] **Summary:** Has a clear description of the high-level functionality and purpose of the software for a diverse, non-specialist audience been provided?
 - [ ] **A statement of need:** Do the authors clearly state what problems the software is designed to solve and who the target audience is?
+- [ ] **State of the field:** Do the authors describe how this software compares to other available packages? Are key relevant or alternative approaches cited and briefly discussed?
+- [ ] **Quality of writing:** Is the paper well written (i.e. such that it does not require editing for structure, language, and writing quality)?
```
kyleniemeyer: minor comment, but `i.e.` needs a comma after it

kyleniemeyer:
While I'm ok removing the paper authors question, I'm hesitant to rely too much on automation, at least based on ORCID—the vast majority of ORCID profiles do not have affiliation listed (anecdotally, based on profiles I've visited), and if there is one it may not exactly match what's in the paper. 

Checking authors & affiliations should probably be an editorial job; I usually ensure that they are complete (i.e., with full details) when I'm doing a final check as AEiC.

I'm also comfortable with modifying the specific question about DOIs, with these caveats:
- whedon's DOI checker is not foolproof, and should only be used as a guide. I don't want authors, reviewers, or editors to assume that they don't have to look at references themselves. As is, I find some people over-relying on this output, for example when it misidentifies something as needing a DOI. 
- So, with that in mind, perhaps the question could be something like "Is the list of references complete, and is everything cited appropriately that should be cited (e.g., papers, datasets, software)?"

arfon:
The rationale for removing the author question is that the paper simply doesn't compile without the authors and affiliations in the `paper.md` so it seems weird to check something like this.

That said, I agree, these do need at least eyeballing but I figure this is part of the editor's (and reviewer's) job when reading the paper.

> I'm also comfortable with modifying the specific question about DOIs, with these caveats:

If you look at the changes proposed in this PR, I don't actually think there are any changes to this now.

> Is the list of references complete, and is everything cited appropriately that should be cited (e.g., papers, datasets, software)?

I actually think this phrasing is better than what we have now and would be happy to add it.

jedbrown:
There are reasons why affiliations on a paper would not match even a well-maintained ORCID, such as when the work was done at a prior institution or as part of a visiting appointment. This is especially prevalent for software in which key authors may have founded the project several years ago. We might want to document this publishing convention because I bet lots of submissions just list the current affiliation.

labarba:
> Is the list of references complete, and is everything cited appropriately that should be cited (e.g., papers, datasets, software)?

I would add here: "and do references in the text use the proper [citation syntax]( https://rmarkdown.rstudio.com/authoring_bibliographies_and_citations.html#citation_syntax)"

... since bad citation syntax is such a common problem!

danielskatz:
> > Is the list of references complete, and is everything cited appropriately that should be cited (e.g., papers, datasets, software)?
> 
> I would add here: "and do references in the text use the proper [citation syntax](https://rmarkdown.rstudio.com/authoring_bibliographies_and_citations.html#citation_syntax)"
> 
> ... since bad citation syntax is such a common problem!

I agree with this, and also think we might want to adjust our sample paper to use some of the citation syntax other than [@abc]

app/views/content/github/_review_checklist.erb
```diff
@@ -25,11 +24,13 @@
...
+- [ ] **Automated tests:** Are there automated tests or manual steps described so that the functionality of the software can be verified?
 - [ ] **Community guidelines:** Are there clear guidelines for third parties wishing to 1) Contribute to the software 2) Report issues or problems with the software 3) Seek support
 
 ### Software paper
 
-- [ ] **Authors:** Does the `paper.md` file include a list of authors with their affiliations?
+- [ ] **Summary:** Has a clear description of the high-level functionality and purpose of the software for a diverse, non-specialist audience been provided?
 - [ ] **A statement of need:** Do the authors clearly state what problems the software is designed to solve and who the target audience is?
-- [ ] **References:** Do all archival references that should have a DOI list one (e.g., papers, datasets, software)?
+- [ ] **State of the field:** Do the authors describe how this software compares to other available packages? Are key relevant or alternative approaches cited and briefly discussed?
```
danielskatz: I'm a bit dissatisfied with this.

For software that was new, this might be a useful point in understanding if people would cite it, but for software that is old and well-used but is not really distinct from other software, what would the authors say?

And I fear this could lead to the author saying something, the reviewer saying "you also should mention X", but still not being complete.

I wonder if we could change it to more simply:

**State of the field:** Do the authors describe how this software compares to other commonly-used packages?

app/views/content/github/_review_checklist.erb
```diff
@@ -25,11 +24,13 @@
...
+- [ ] **Quality of writing:** Is the paper well written (i.e., such that it does not require editing for structure, language, and writing quality)?
```
danielskatz: The grammar here is a bit awkward - better would be:

**Quality of writing:** Is the paper well written (i.e., it does not require editing for structure, language, or writing quality)?

kyleniemeyer:
@danielskatz regarding citation syntax, in the `# Citations` section we do now include the different formats authors can use (`@author:2001`, `[@author:2001]`, `[@author:2001; @author2:2010]`).

danielskatz:
Thanks - you are correct - I hadn't read this lately.

arfon:
Fixes https://github.com/openjournals/joss/issues/442

app/views/content/github/_review_checklist.erb
arfon: What do you think @Kevin-Mattheus-Moerman?

Kevin-Mattheus-Moerman: I see what @danielskatz is saying. His proposed shorter/simpler form sounds fine too.

Issue #531, opened 2019-05-04
Feature suggestion, additional search mask


nbehrnd:
New to the journal, I started to read some papers filed as [accepted](https://joss.theoj.org/papers/accepted).  Currently (2019-May-04), this category applies to 556 of 663 papers there, spread over 56 pages.

To facilitate search (and eventually, equally retrieval) of the publications of interest, I would like to suggest an additional interface, allowing to identify the articles of interest by
- author(s) name
- keywords extracted e.g. from the short abstract on this "directories page" by JOSS, and a subject where the program may find application (e.g., chemistry, crystallography)
- similar to Clarivate's database 'Web of Science', a retrieval by the doi eventually attributed
Relating to github, either an indicator or/and an additonal search criterion might be the language used (e.g., Fortran, Python, R).

ivan-marroquin:
Hi,

As pointed out by @nbehrnd , it will be a great addition to JOSS by having a search capability. 

Thanks to the team for making this journal available to us!

Ivan

anmolter:
Hi,
I also think that this would be very useful. I tried to search for JOSS articles in the standard journal search engines (e.g. Google Scholar, Pubmed), but they don't retrieve anything, even if I use words from the title of an accepted article. 
Anna

nbehrnd:
Hi Anna,

/perhaps/ OpenAire is a better tool to retrieve publications in JOSS
for which title and author already are known.

As by today (2019-Jun-04), the latest accepted article is «ropensci /
cde - R package to retrieve data from the Environment Agency Catchment
Data Explorer site» by Robert A. Briers.  It is noticed that its
top-site [1] attributes as doi 10.21105/joss .01473, different to the
one zenodo attributes (10.5281/zenodo.2577178) which then equally is
the one used in openAire [2].  The more general search mask by OpenAire
seems to cover more than publications though [3].

As for the different DOIs, which may be a pitfall for literature
management programs (e.g., zotero), at present, the doi-resolver [4]
accepts the zenodo DOI, but not yet "10.21105/joss.01473" (without the
quotation marks).  Perhaps this due to a delayed update of the
corresponding data bases which will resolve automatically in the
course of some hours / days.

nbehrnd

[1] https://joss.theoj.org/papers/0d35f75e861fcf47556d70571e226589
[2]
https://explore.openaire.eu/search/software?softwareId=r37b0ad08687::8cf0c797072ceca9c47c6dd19a54f265
[...]

PR #552, opened 2019-06-26
[WIP]: New JOSS designs

arfon:
This PR is a work in progress. The test version of the JOSS site is available here: http://staging.joss.theoj.org/

At this point I've not tested a lot of the site, feel free to click around and highlight any issues you find @openjournals/joss-editors. Also, it's probably *not* a good idea to try and use this site for doing anything real at this point (e.g. submitting papers).

Known issues at this point (please comment with bugs/issues you find):

- [ ] Dashboard views are incomplete (although they do work)
- [x] RSS links for papers can't be clicked on.


kyleniemeyer:
This is looking great! 

A few comments/suggestions for consideration:
- in the editors dashboard, can the emeritus editors be shown separately or at the bottom? it might also be nice to either have the names listed in ascending order of current assignments, although alphabetical is good for quickly finding someone. Maybe a future upgrade is to make the rows sortable based on one of the columns?
- A search feature on the papers list would be very helpful. Right now it's only possible to find a paper by going through one-by-one; in the long term, using tags in some way would be great, but just a basic search would be nice for now

karthik:
The design looks great! The avatars for the editors are squished due to aspect ratio problems.

karthik:
> A search feature on the papers list would be very helpful. Right now it's only possible to find a paper by going through one-by-one; in the long term, using tags in some way would be great, but just a basic search would be nice for now

Search is great! I wonder if we can ask for keywords in the submission template and make papers searchable that way too.

kyleniemeyer:
> I wonder if we can ask for keywords in the submission template and make papers searchable that way too.

Most papers, particularly those following the example we have in the docs, already do include keywords.

karthik:
Cool. Then we just need to add those as tags/labels to whatever is generating the static site.

karthik:
Cite As could have some some of UI that allows for copying the citation in bib format.

danielskatz:
On the front page, we list each paper with the author indicated by the GitHub handle of the submitter.  I feel like we should actually list the full names of all the authors that the submitter submitted.

danielskatz:
the new page links for the papers look nice - I'm glad to see the altmetrics badge

danielskatz:
On the page for the paper, where we give the citation information, how do we decide to use et al.?  I would think with that with 3 or less authors, we would actually specify the authors, and perhaps with more than 3, use et al.

And when we do use et al., why do we only provide the surname of the first author, and not their initials?

danielskatz:
on the about page, I think the heading "Editors Emeritus" should be "Emeritus Editors"

danielskatz:
The blog could probably use some type of table of contents to show the posts that there are there, rather than having to scroll through the whole thing to find them

danielskatz:
On the front page, "view editorial guidelines" goes to read the docs, but it goes to "reviewing for JOSS", not "Editorial Guide" - I think the link on the front page should be renamed "reviewer guidance"

danielskatz:
Does the info in the my profile tab need any wording for GDPR?

danielskatz:
> in the editors dashboard, can the emeritus editors be shown separately or at the bottom?

seconded - I think I've requested this previously as well, perhaps via an issue?

danielskatz:
In the dashboard/stats, there's something off about the editors - the first editor listed doesn't seem to have a name. Similarly, in the Accepted papers per month by editor, the pulldown has a blank entry in the middle.

danielskatz:
Overall, this looks like a nice set of changes.

labarba:
- The list of papers is labeled "Accepted Papers" and the papers have an "Accepted" label. Shouldn't this be *Published Papers* instead?

> on the about page, I think the heading "Editors Emeritus" should be "Emeritus Editors"

- No: _emeritus_ is a postpositive adjective: it is placed after the word it modifies.

danielskatz:
> The list of papers is labeled "Accepted Papers" and the papers have an "Accepted" label. Shouldn't this be Published Papers instead?

agreed

danielskatz:
>> on the about page, I think the heading "Editors Emeritus" should be "Emeritus Editors"

> No: emeritus is a postpositive adjective: it is placed after the word it modifies.

Sorry, I don't agree that this is how it is commonly used in English, though it's certainly possible there's a rule that says it should be.

labarba:
- On the [papers](http://staging.joss.theoj.org/papers) view, I think the "Accepted papers" (which I submit should be "Published") tab should be the right-most (and default) tab, followed by "Active" and "All" to the left.

Kevin-Mattheus-Moerman:
Below are my comments on the site (mainly on the mobile version) with screen shots. I think it would be **AMAZING** if we rendered (optional provided on submission) graphical abstracts with the papers. Here is an example of what that could look like: 

joosPage

# Desktop site (large screens


# Mobile site (small screens)

## The paper(s) view








Can the PDF be rendered for the mobile version too? 

## About and editor view

lheagy:
This is looking beautiful!! I have a couple minor suggestions:

- On the main, page, I think it would be helpful to have a "submit" and "volunteer to review" in the body as well, either with the "explore / documentation / etc" or right under the short description of JOSS as these are two main actions we want people to take to engage with us

image


- The bouncing JOSS logo at the bottom of the page is fun, but I find it a bit distracting (sorry to be the kill-joy...) and I also expected to be able to interact with it (e.g. I thought clicking on it would do something - as usually things that have some sort of motion, at least on my mac, usually mean they want my attention for an update or something) 

- the might be out-of-scope for the current re-vamp, but it would be nice to have the ability to search for a paper from the paper listing
image

arfon:
> fix links to papers' RSS feeds cb06059

Thanks @xuanxu. If you want to try and debug why the links in these drop downs aren't actually clickable that would be awesome 😬

app/views/content/layout/_navbar.html.erb
```diff
@@ -25,7 +25,7 @@
       <%- end %>
 
       <%= active_link_to "About", about_path, class: "nav-item nav-link" %>
-      <%= active_link_to "Papers", papers_path, class: "nav-item nav-link" %>
+      <%= active_link_to "Papers", published_papers_path, class: "nav-item nav-link" %>
       <%= active_link_to "Docs", "https://joss.readthedocs.io/en/latest/index.html", target: '_blank', class: "nav-item nav-link" %>
```
xuanxu: I was going to add this change to `views/papers/show.html.erb` to keep the top links behaving the same way everywhere but that raised another question: why is the paper' show view not using the navbar partial? 
Just in case I'm missing something I've issued this change as a separate [PR against the fresh branch here](https://github.com/openjournals/joss/pull/555).

arfon: Thanks for that PR @xuanxu!

labarba:
On the paper view page, I feel like the legend "View Review" is awkward. How about "Paper Review" instead?
Screen Shot 2019-07-18 at 3 16 40 PM

arfon:
Hi all, apologies for the slow reply here. Thanks for all of the excellent comments! Many of the comments and suggestions that you've made have been incorporated ✨ 

> A search feature on the papers list would be very helpful. Right now it's only possible to find a paper by going through one-by-one; in the long term, using tags in some way would be great, but just a basic search would be nice for now

Yes, this is in the works but is going to require some additional effort. This is essentially the next item on the development roadmap but isn't going to make it in as part of this PR / cc #531 #369 

> On the front page, we list each paper with the author indicated by the GitHub handle of the submitter. I feel like we should actually list the full names of all the authors that the submitter submitted.

Good idea. There's some additional work required in Whedon for this to function which is partly finished in https://github.com/openjournals/whedon/pull/40. Again, I won't get this change fixed in this PR but it's in the backlog.

> Below are my comments on the site (mainly on the mobile version) with screen shots. I think it would be AMAZING if we rendered (optional provided on submission) graphical abstracts with the papers. Here is an example of what that could look like:

I do like this idea, but I still don't have a good way to handle this. Perhaps we could have an optional YAML header key for the paper that would allow someone to point us at a graphic?

At this point I think we've got a great set of changes which dramatically improve the current site so I'm going to go ahead and merge this. Further updates to address some of the bigger items discussed here will be happening over the next couple of months.

PR #556, opened 2019-07-08
adding whedon generate pdf example to pre-review issue

danielskatz:


arfon:
Good idea. Unfortunately this is in the wrong place - the edits you've made here are to some fixtures for the tests. I've added this language over in https://github.com/openjournals/whedon-api/commit/6611b88327b3dc0e19af97b12a6ee0a68b8ae90d

danielskatz:
thanks - the source tree is extremely confusing to me, so I just grep for what I want to change and then change it - maybe I should have paid more attention to multiple results :)

arfon:
> thanks - the source tree is extremely confusing to me, so I just grep for what I want to change and then change it - maybe I should have paid more attention to multiple results :)

Totally reasonable strategy :-)

danielskatz:
although in this case, where the change I needed to make was in a different repo, I don't think I could have been successful

PR #559, opened 2019-07-12
Removes specific author username from documentation

arokem:
Looks like maybe this was a copy-paste?


arfon:
:+1: good catch @arokem

PR #566, opened 2019-07-26
clarify steps after submission

nuest:
This might seem like too much hand-holding, but I think it's worth avoiding impatient folks like me expecting to see something on GitHub seconds after the submission.


arfon:
Hi @nuest, many thanks for submitting this. In #567 I believe I've addressed this, specifically:

- Submitting authors now receive an email when they submit telling them what the next steps are
- Freshly-submitted papers now show up properly on the user profile page.

As such, I'm going to close this PR but many thanks for opening it! It prompted me to work on this :-)

PR #580, opened 2019-08-13
Rename ISSN setting

xuanxu:
`twitter_issn` -> `issn`
It makes no sense for the ISSN setting to have `twitter` in its name.


arfon:
😆agreed, this is nonsense. Thanks for fixing this @xuanxu.

PR #581, opened 2019-08-13
Update deprecated method

xuanxu:
`success?` has been deprecated in favor of `successful?`
This PR removes deprecation warnings from the test builds.


arfon:
💖

PR #584, opened 2019-08-14
Show complete title of paper on mouse over

xuanxu:
The title of the paper is truncated in the dashboard tables but in some cases I want to read the complete title without clicking on it. This PR just adds the title of the paper to the link so it shows when mouse is over it:

title


arfon:
👍 good idea.

PR #588, opened 2019-08-18
removing 0000 ORCID from example paper

danielskatz:


danielskatz:
I'm fairly sure this is right...

arfon:
👍 thanks.

Issue #130, opened 2016-06-03
Google Scholar crawling 


genomematt:
It does not at the moment look like JOSS is being indexed by google scholar.
I think this is one of the sites we want to ensure visibility for the journal on.

https://scholar.google.com.au/intl/en/scholar/inclusion.html#troubleshooting

arfon:
JOSS papers have included the Google Scholar metadata tags since [this update](https://github.com/openjournals/whedon/commit/4784cff8aec87cc3e10d454294bf06a02f8f6a10) about 10 days ago.

Their docs suggest it can take 4-6 weeks to be indexed so I think we're still in that window. Let's leave this open for now to keep tracking this issue. I agree we must be indexed by Google Scholar, I'm just not sure there's anything we need to do other than wait at this point 🕐

pjotrp:
Excellent!

sherrillmix:
At least a couple papers show up now:
https://scholar.google.com/scholar?hl=en&q=GeneNetwork%3A+framework+for+web-based+genetics
https://scholar.google.com/scholar?hl=en&q=R3D2%3A+Relativistic+Reactive+Riemann+problem+solver+for+Deflagrations+and+Detonations

However other articles on the same page currently do not:
https://scholar.google.com/scholar?q=Xenomapper%3A+Mapping+reads+in+a+mixed+species+context
https://scholar.google.com/scholar?q=pyuca%3A+a+Python+implementation+of+the+Unicode+Collation+Algorithm

Pretty cool that some are showing anyway and maybe google's getting around to processing the others.

arfon:
@sherrillmix, thanks for finding these. Let's keep monitoring this.

FaustinCarter:
It looks like Google Scholar is not actually indexing anything from JOSS. The two examples listed by @sherrillmix only show up in Google Scholar because the first is also listed at researchgate.net and the second is listed at eprints.soton.ac.uk. Google Scholar is indexing them based off those domains, not because of anything going on at JOSS.

cMadan:
Maybe Google Scholar isn't crawling through the Github links? My JOSS paper is indexed, but that's from being uploaded to my own website (https://scholar.google.com/citations?view_op=view_citation&citation_for_view=2QJwoAwAAAAJ:3BvdIg-l-ZAC), but I feel like that's still sufficiently different than being on researchgate or a university's own repository.

pjotrp:
[...]

PR #605, opened 2019-09-16
Modifying PDF URLs for papers to match Google Scholar redirects

arfon:
This PR changes the JOSS web application to point to PDFs local to the JOSS domain from the JOSS UI. This is a requirement for Google Scholar indexing:

> The PDF file for the article should be on the same host as the page with the metatags. Highlighted in green above. If this is fixed, would be best to update the citation_pdf_url metatag as well

/ cc #130


app/controllers/papers_controller.rb
```diff
@@ -196,7 +196,10 @@ def show
       head 404 and return unless can_see_hidden_paper?(@paper)
     end
 
-    render layout: false
+    respond_to do |format|
+      format.pdf { redirect_to @paper.pdf_url, :status => 301 }
```
arfon: If format is `pdf` we simply redirect to the existing `pdf_url` on GitHub pages.

app/models/paper.rb
```diff
@@ -255,6 +255,26 @@ def joss_id
     "#{setting(:abbreviation).downcase}.#{id}"
   end
 
+  # This URL returns the 'DOI optimized' representation of a URL for a paper
+  # e.g. https://joss.theoj.org/papers/10.21105/joss.01632 rather than something
+  # with the SHA e.g. https://joss.theoj.org/papers/5e290cb57b61f83de4460fd0eca22726
+  # This URL format only works for accepted papers so falls back to the SHA
+  # version if no DOI is set.
+  def seo_url
```
arfon: There's probably a better way to do this with Rails routes and custom paths.

arfon:
:wave: @xuanxu - this change works fine in Chrome and Firefox but Safari is giving me some kind of CORS error - perhaps it doesn't like the `301` for the PDF now?

Any ideas what might be going on here?

xuanxu:
@arfon :thinking: I can't think of a reason for only failing in Safari. I'll try to test later today.

arfon:
This feels related: https://github.com/NUKnightLab/StoryMapJS/issues/133

arfon:
I just had the realization that I think the only URL that needs to point to the `https://joss.theo.org/.../blah.pdf` host is the Google Scholar meta tag (e.g. [this line](https://github.com/openjournals/joss/pull/605/files#diff-d426aa85f1b9364c3828995435a6aba7R62)) which has nothing to do with the embed view for the PDF.

So, I think I'm going to say this is fixed and that this is good enough for now.

sayanddude:
@arfon Hi! I just had one question regarding indexing of my paper- https://joss.theoj.org/papers/10.21105/joss.01509 in Google scholar. The paper was published around 2 month back but it's not yet indexed in Google scholar whereas I see some recently published ones indexed in Google Scholar. Is there any way to fix this?

arfon:
Good question! I'm not sure to be honest. The Google Scholar bot _eventually_ seems to pick up papers, why yours has not yet been indexed is a mystery sorry...

sayanddude:
> Good question! I'm not sure to be honest. The Google Scholar bot _eventually_ seems to pick up papers, why yours has not yet been indexed is a mystery sorry...

Ok.... This is really strange!

PR #625, opened 2019-11-05
Papers by reviewer and editor

xuanxu:
This PR adds a `reviewed by` filter to Papers. 

And links that list from the paper show view:
Screenshot_2019-11-05 AyComPy


arfon:
Looks great, thanks @xuanxu. I figured we should add 'filter by editor too' while we were at it...

Issue #623, opened 2019-11-02
JOSS website should have a clear "donate" button/link


danielskatz:
I was thinking of tweeting that people should consider donating to JOSS, but our front page doesn't have button or line - we are good at putting "volunteer to review" buttons around, but not donate...

Also, the about page has a cost section that talks about donations, but no way to donate from there.

danielskatz:
#627 was a partial solution, but as I wrote there, can we also add a more prominent "donate" link, perhaps in the right column index/toc on the about page?  (the small badge is very small and very hidden down at the bottom)

arfon:
OK, I just tried to improve this in https://github.com/openjournals/joss/commit/e882afbb5c472d9ea9b931cc98d05d1e0f639679

<img width="1484" alt="Screen_Shot_2019-12-11_at_6_22_07_PM" src="https://user-images.githubusercontent.com/4483/70669044-4426c700-1c43-11ea-89a6-094fa94d789c.png">

danielskatz:
thanks magic issue genie - this looks good!

I'm happy for that to be merged and this issue to be closed

PR #627, opened 2019-11-08
Add 'donate' links and badges

xuanxu:
This PR adds "donate" links to the site's `About` page and to the repo's `README` page.

Re: #623


danielskatz:
The title is a bit misleading - I think this is for JOSS's page, not NumFOCUS's.  Otherwise, this looks reasonable, though it's a bit hard to see without actually being in context in the live (or a test) site.

xuanxu:
Here's a screenshot, the PR just adds a link and a badge to the cost section:

Before:
before

After:
after

danielskatz:
that looks good, thanks.

Could we also create a donate word, button, or link in the right side "TOC" in the about page?

danielskatz:
@arfon - can you merge this one too, since I think it's much more likely that this is where potential donors will come, rather than to the github page

Issue #641, opened 2019-12-04
Suggestion for JOSS: metadata for Zotero


psmbln:
Hi everyone,

I am using Zotero as a reference manager:
https://www.zotero.org/

Zotero tries to import embedded metadata from article landing pages, but this doesn't seem to work very well yet for JOSS articles. Please have a look into this; it would be great for a software journal to interact seamlessly with bibliographic tools.

Thanks and kind regards,
Paul Schultze-Motel

psmbln:
more specifically:
- Item Type is recognised as "Web Page" (should be: "Journal Article") -> this leads to missing entries in fields 
  -"ISSN"
  - "Volume"
  - "Issue"
  - "Page"
  - DOI is mapped to "Extra" (instead of "DOI")
- URL is recognised incompletely

arfon:
Thanks for reporting this @psmbln. Can you take a look again now that we've updated the Item Type in #642?

psmbln:
Thanks - I have now tested the metadata import to Zotero with a recent JOSS article (cf. attached screenshot).

![driscoll2019_zotero](https://user-images.githubusercontent.com/29654660/70308771-47b5ea80-180c-11ea-8b78-29e7439abbba.png)

The issues above are now fixed, exept:
- URL is recognised incompletely

One more thing:
- the abstract is possibly not recognised correctly; in the "Abstract" field, there is instead a citation of the article

arfon:
Thanks for helping us debug this @psmbln!

> URL is recognised incompletely

This should be fixed now. Do you think you could check again?

> the abstract is possibly not recognised correctly; in the "Abstract" field, there is instead a citation of the article

Right. This is going to be an issue as JOSS papers don't have abstracts. I'm not sure what we can do about this one...

psmbln:
Great, thank you - the complete URL is now recognised by Zotero.

About the abstracts: if there is no abstract, would it be possible to leave the "Abstract" field empty?

arfon:
> About the abstracts: if there is no abstract, would it be possible to leave the "Abstract" field empty?

I'm guessing it's picking up the `meta property="og:description"` field in the HTML and using that as the content for the abstract. 

```
<meta property="og:description" content="Driscoll, (2019). ComplexRegions.jl: A Julia package for regions in the complex plane. Journal of Open Source Software, 4(44), 1811, https://doi.org/10.21105/joss.01811" />
```

[...]

PR #647, opened 2019-12-11
Add paper url to meta tags

xuanxu:
Thi PR adds the seo url to the paper meta tags, so DOI for published papers, SHA otherwise
re: #641


xuanxu:
@arfon: ouch, I've just realized [you changed this](https://github.com/openjournals/joss/commit/ed2ccca2a696b425b0b759545f52dcf6f2cacd02).
I'm keeping the PR open because I think this is a more complete solution (it changes the url in more places).

arfon:
👍 thanks for this @xuanxu, your fix is better than my hack 😸

PR #651, opened 2019-12-16
Rails version up to 6.0.2

xuanxu:
we're so edge


arfon:
> we're so edge

🔥 

@xuanxu - any idea how to fix the [build error](https://travis-ci.org/openjournals/joss/builds/625634294?utm_source=github_status&utm_medium=notification) with Travis?

xuanxu:
@arfon there's a new version of Bundler [released two days ago](https://rubygems.org/gems/bundler) and it's causing some kind of conflict. 
Pushing a new `Gemfile.lock`, let's try to use the new version...

arfon:
✨ thanks!

Issue #637, opened 2019-11-29
google scholar indexing


jcolomb:
hi,
I have seen that a JATSxml version is necessary for pubmed (and maybe also base), is it also the case for google scholar or can we do something there?

arfon:
Hi @jcolomb - we've been tracking Scholar indexing over in https://github.com/openjournals/joss/issues/130. Please see that thread for context.

PR #656, opened 2019-12-27
Updated logic for PDF responses

arfon:
This is an update to the controller logic for handling PDF responses. In the past we redirected PDF urls but it turns out that the Google Scholar webcrawler doesn't like this. 

With this change, we download the file server-side and send directly to the browser when requesting a PDF, rather than redirecting.

/ cc #130 #637 @labarba


spec/controllers/papers_controller_spec.rb
```diff
@@ -201,11 +201,20 @@
...
 
-      get :show, params: {doi: paper.doi}, format: "pdf"
-      expect(response).to redirect_to(paper.pdf_url)
+      get :show, params: {doi: paper.doi}
+      expect(response.body).to eq(IO.binread('spec/fixtures/paper.pdf'))
+    end
+
+    it "should not redirect when accepting any content type" do
+      paper = create(:accepted_paper)
+      request.headers["HTTP_ACCEPT"] = "*/*"
```
arfon: This is the content type the Google web crawler specifies.

Opussci:
The inclusion guidelines for the Google Scholar webcrawler state that abstract must be visible on the article page. As I see your journals do not display it.
Maybe that could resolve indexing issue if it is not already resolved.

arfon:
> Maybe that could resolve indexing issue if it is not already resolved.

Thanks @Opussci. Our indexing issues are actually fixed now.

Opussci:
Then sorry for spamming!

PR #662, opened 2020-01-09
API endpoint to reject papers

arfon:
This is a small addition to the JOSS API to support rejecting papers from Whedon.


config/routes.rb
```diff
@@ -52,6 +52,7 @@
   post '/papers/api_deposit', to: 'dispatch#api_deposit'
   post '/papers/api_assign_editor', to: 'dispatch#api_assign_editor'
   post '/papers/api_assign_reviewers', to: 'dispatch#api_assign_reviewers'
+  put '/papers/api_reject', to: 'dispatch#api_reject'
```
arfon: Someday soon we should actually refactor this to be a proper RESTful API...

PR #665, opened 2020-01-12
Updating suggested editor flow

arfon:
This is a small update to the submission flow when authors suggest the handling editor on the JOSS submission form. Currently an EiC has to decide between:

1. Accepting the author's suggestion and then automatically assigning the JOSS editor before they have had a chance to say yes/no?
2. Ignoring the author suggestion.

With this update, suggestions by submitting authors are _just that_. Even if the EiC keeps this suggestion on the JOSS dashboard, it will only show up in the reviews repository as a small amount of additional language from Whedon:

```
:wave: @joss_editor would you be able to handle this submission for JOSS?
```

Screen Shot 2020-01-12 at 11 56 21 AM

@openjournals/joss-eics

### DIFF
--- a/app/models/paper.rb
+++ b/app/models/paper.rb
@@ -332,9 +332,9 @@ class Paper < ActiveRecord::Base
 
   def meta_review_body(editor)
     if editor.strip.empty?
-      locals = { paper: self, editor: "Pending" }
+      locals = { paper: self, suggested_editor: "Pending" }
     else
-      locals = { paper: self, editor: "#{editor}" }
+      locals = { paper: self, suggested_editor: "#{editor}" }
     end
     ApplicationController.render(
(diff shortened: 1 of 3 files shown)


labarba:
This change aligns the behavior of whedon to the workflow already adopted by EiCs, which is to bypass automatic assignment of a handling editor, and ask for the editor's agreement—after considering not only topic of the submission but also editor load. 

Because EiCs invite handling editors taking into account their load, and possibly if they are `ooo`, I would leave the Pre-Review language simply as "The author's suggestion for handling editor is `@user`".

danielskatz:
I agree with @labarba's suggestion, but otherwise this looks good to me

arfon:
Thanks for the feedback @labarba and @danielskatz!

Issue #315, opened 2017-08-22
Add browse/search options for papers


nicoguaro:
I think that it would be useful to have the option to search in the Journal

arfon:
WIP solution to this in https://github.com/openjournals/joss/pull/369

Miserlou:
+1 for search.

cdeil:
Would be really nice to have a way to see all the papers.
Even a single HTML page would to, then I could use the page search functionality in my browser to find some author or keyword.

pllim:
I was just looking for this exact feature.

Fun fact: Searching on Google landed me at https://en.wikipedia.org/wiki/Joss_paper .

Fil:
vote up

banesullivan:
Is there a status update on this? Or even a workaround solution if I'd like to search for papers on different topics?

jordibc:
A workaround: https://www.google.com/search?q=site%3Ajoss.theoj.org+YOURSEARCH

It would be so much better with a native searching functionality. Looking forward to having it!

banesullivan:
Thanks @jordibc!!

karllark:
would be great to have this search capability.  Adding my voice to this request.

PR #683, opened 2020-02-15
Adding search box for papers#index

arfon:
This PR adds a search box on the paper listing pages to allow readers to search paper titles, author-defined tags, and programming languages associated with the software.

Screen_Shot_2020-02-15_at_8_24_23_AM

As before, it's still possible to filter papers by a specific tag by clicking these in the UI.

Fixes https://github.com/openjournals/joss/issues/315


Fil:
It's great! I was a bit surprised that so many papers were returned by the "d3" query https://joss.theoj.org/papers/search?q=d3 ; for some reason it seems that those papers are about "3D" more than d3.js (except for the two first papers, which are indeed using d3.js).

arfon:
I think this might be the fuzzy string matching and the fact that `d3` is such a short string, i.e. `d3` is one edit away from `3d`.

PR #692, opened 2020-02-27
Jed/submitting with refs

jedbrown:
I observe a lot of diverging styles and manual referencing so I think it's useful for the example to clean and consistent ways of doing this. This PR is the best recommendation I know, but I'd be happy to adjust for other conventions.

### DIFF
--- a/docs/submitting.md
+++ b/docs/submitting.md
@@ -156,6 +156,12 @@ Figures can be included like this:
 ![Caption for example figure.\label{fig:example}](figure.png)
 and referenced from text using \autoref{fig:example}.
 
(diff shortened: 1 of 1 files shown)


jedbrown:
@arfon Does Travis always fail for PRs from external repositories? The errors don't look like they'd have anything to do with my change.

arfon:
> @arfon Does Travis always fail for PRs from external repositories? The errors don't look like they'd have anything to do with my change.

Yeah, there's an environment variable/secret in the Travis config which doesn't seem to be picked up properly on external PRs.

Thanks for this change - very useful!

PR #695, opened 2020-02-29
Dashboard refresh

arfon:
:wave: @openjournals/joss-editors @openjournals/joss-eics 

This PR brings the editor dashboard and summary screens up to date with the latest site designs. It also adds a little extra information to the dashboard views for each editor:

dashboard

The new information available is as follows:

- Number of papers published in the last month that you edited.
- A rolling 3-month average for the number of papers being published per month edited by you.
- A rolling 3-month average for the number of papers being published per editor per month across all of JOSS.

Obviously for newer editors, these numbers may be zeroes, especially as papers take some time to process 😄.

Please let me know if you find any weird quirks with these numbers or if you have suggestions for more informative metrics. I think I did my maths correctly here 🤞

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,7 +21,7 @@
 # Ignore readthedocs build
 
 docs/_build
-latest.dump
+latest.dump*
 
 /public/packs
 /public/packs-test
--- a/app/assets/stylesheets/home.scss
+++ b/app/assets/stylesheets/home.scss
@@ -16,8 +16,48 @@
   }
 }
 
+.my-statistics {
+
+  .col {
+    padding-right: 0px;
+  }
+
+  .banner {
+    border-bottom: 1px solid rgba(46, 41, 78, 0.5);
+  }
+
+  span {
+    font-size: 0.9em;
+    font-weight: bold;
+    margin-bottom: 0;
+
+    a {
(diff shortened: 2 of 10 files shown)


labarba:
I'm a little worried that comparing the immediate-past 3-month rolling average for the editor, with the all-time monthly average for all editors, may send the wrong signals. The load on editors has changed over time, no? Why not use the same time periods for these two averages? (And 3 month could be too little: maybe 6?)

arfon:
Sorry, just to be clear, both of these numbers are monthly averages across the last three months.

- The per-editor number (middle column) is the number of papers per month, averaged over a three month period. So in my case, 9 papers have been published in JOSS over the last three months which makes my monthly average 3.0.
- The 'all JOSS' last metric is the number of papers per month per editor across the whole of JOSS _in the last three months_ too. So the time periods are actually the same.

Does that make more sense?

oliviaguest:
Sorry if this is the wrong place to ask, but is there a way to see all papers I have edited (not just ones that have not been accepted yet)? 😅

labarba:
@oliviaguest Go to any published paper you edited, and look at "Editor" on the right-hand side, there's an "all papers" link.

PR #711, opened 2020-03-13
Skip specs affected by closing submissions

xuanxu:
This PR refactors the disabling of tests caused by suspending new submissions, so it will be easier to find when reverting to normal situation.
- papers#create route is disabled
- affected tests are skipped with proper message:

rake spec


arfon:
👍 thanks @xuanxu!

PR #713, opened 2020-03-15
Make search for author/editor/reviewer not exact

xuanxu:
Since [this change](https://github.com/openjournals/joss/commit/f82479e6359bc6a87e1c77bca00827d5c1843a16#diff-80c2de5496f8beac6c1754149c0a30f4L110-R124) some searches are not working. Esamples:

- Empty 'by author' seach: if I click in any of the author's link in this recently published paper:  https://joss.theoj.org/papers/10.21105/joss.02102 the resulting page for "All papers by XXX" does not return any [result](https://joss.theoj.org/papers/by/Samantha%20Hood).
- By reviewer/editor searches: These [editor](https://joss.theoj.org/papers/edited_by/@arfon)/[reviewer](https://joss.theoj.org/papers/reviewed_by/@arfon) work, but [these](https://joss.theoj.org/papers/edited_by/Arfon) [two](https://joss.theoj.org/papers/reviewed_by/arfon) don't. I think they should return the same results.

This PR just removes the `exact` search attribute for the `:authors`, `:reviewers` and `:editors` fields.

### DIFF
--- a/app/controllers/papers_controller.rb
+++ b/app/controllers/papers_controller.rb
@@ -107,21 +107,21 @@ class PapersController < ApplicationController
       @term = "in #{params['language']}"
 
     elsif params['author']
-      @papers = Paper.search(params['author'], fields: [authors: :exact], order: { accepted_at: :desc },
+      @papers = Paper.search(params['author'], fields: [:authors], order: { accepted_at: :desc },
                   page: params[:page],
                   per_page: 10
                 )
       @term = "by #{params['author']}"
 
     elsif params['editor']
-      @papers = Paper.search(params['editor'], fields: [editor: :exact], order: { accepted_at: :desc },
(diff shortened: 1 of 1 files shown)


arfon:
Thanks @xuanxu. I made this change without much testing :-\ (can you tell? 😸). I'd still like to make some improvements on these filters, especially because _without_ the `exact` matches then searches like this one https://joss.theoj.org/papers/by/Arfon find papers where the author name includes `Aron`.

It's not exactly clear how to solve both sets of problems here, I suspect this could be addressed by better understanding of Elastic Search and SearchKick :-)

xuanxu:
> _without_ the `exact` matches then searches like this one https://joss.theoj.org/papers/by/Arfon find papers where the author name includes `Aron`.

@arfon I think we can fix that disabling misspelling. PR incoming!

PR #725, opened 2020-04-23
Removing inline render

arfon:
EOM


arfon:
It doesn't work unfortunately :-(

arfon:
Can't get this to work, but I'm not sure the RCE is a reasonable concern given that we're not accepting any params to render the SVG.

PR #738, opened 2020-05-23
Fix typo

Nosferican:
Rather than OSD it should be OSI.


arfon:
Thanks for spotting this @Nosferican!

Issue #709, opened 2020-03-12
Flag borderline reject submissions that move to Review


labarba:
1. Our process of open peer reviews is incompatible with giving reviewers the decision power to reject submissions. The accept/reject decisions should always be made by the editors.

2. Most of the time, we would like to flag out-of-scope or minor submissions in the **Pre-Review** stage, as this is much cleaner than doing it after it goes to Review. But, in some circumstances, the deeper inspection made by reviewers is needed to make a determination.

3. “Quantity of code” is not a universal measure, and we can’t use it blindly to reject submissions as minor. But _most of the time_ it is a good measure. We need a bar below which all submissions are inspected for being possibly “minor.” Say, 1,000 lines of code. It doesn’t mean we reject if it doesn’t pass that bar, it means that research use and scholarship content needs to stand out for submissions under this bar to move forward.

4. Submissions that were borderline and discussed in the Pre-Review stage as possibly out-of-scope or minor should be _flagged_ as such if they move to Review. This can ease the burden on reviewers: we tell them, look, we think this should be rejected, but are giving it a chance to prove research need: can you find reasons to defend it? So the reviewer is not asked to reject; they have to be enthusiastic for us to “revert” our Pre-Review assessment to reject. If they are just indifferent, the Pre-Review assessment stands.

arfon:
Thanks for summarizing this @labarba. 

> We need a bar below which all submissions are inspected for being possibly “minor.” Say, 1,000 lines of code.

I think this is a reasonable level to trigger some kind of default process such as the handling editor labelling the issue as `flagged as minor` or something similar. (It would be great if we could figure out how to get Whedon to do this too).

> It doesn’t mean we reject if it doesn’t pass that bar, it means that research use and scholarship content needs to stand out for submissions under this bar to move forward.

👍👍 👍  

> So the reviewer is not asked to reject; they have to be enthusiastic for us to “revert” our Pre-Review assessment to reject. If they are just indifferent, the Pre-Review assessment stands.

This definitely sounds like it could potentially streamline our decision-making process as an editorial team, especially if the semantics of having the `flagged as minor` label means that unless demonstrated otherwise, the submission is assumed to be rejected post-review unless good reasons are found to defend it by reviewers.

One question I have based on this proposal:

- Do you think we need one or two labels here? We want to both flag an incoming submission as potentially a minor submission, plus we want to be able to denote something that was discussed at pre-review stage and deemed possibly out of scope. Are these two the same? (Possibly not).

[...]

PR #764, opened 2020-07-01
Stubbing out code for editor votes

arfon:
This PR implements a possible way for the JOSS editorial team to make decisions about editorial scope. Our [current approach](https://joss.readthedocs.io/en/latest/editing.html#rejecting-a-paper) for flagging/discussing whether a paper is in scope is very noisy for editors (messages posted to the Google group), hard to keep track of, and prone to errors.

The idea is that any `pre-review` or `review` issue labeled with `query-scope` will automatically be included in the twice-weekly emails to all editors to prompt them to vote.

**Specifically this PR implements**

A UI addition to provide a way for editors to vote on whether or not a submission is in scope:

Screen_Shot_2020-07-01_at_1_30_39_PM

Vote summaries added to the dashboard views:

Screen_Shot_2020-07-01_at_1_29_49_PM

A summary email that is sent to editors on Mondays and Thursdays:

Screen_Shot_2020-07-01_at_1_31_35_PM


app/controllers/votes_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    params = vote_params.merge!(editor_id: current_user.editor.id, kind: kind)
+
+    @vote = @paper.votes.build(params)
+
+    begin
+      @vote.save!
+      flash[:notice] = "Vote recorded"
+      redirect_to paper_path(@paper)
+    # Can't vote on the same item twice
+    rescue ActiveRecord::RecordNotUnique => e
```
arfon, line 23: Each editor can only vote on each paper once 😬

app/helpers/home_helper.rb
```diff
@@ -37,6 +37,9 @@ def pretty_labels_for(paper)
           concat content_tag(:span, label, style: "padding: 3px; margin-right: 3px; border-radius: 2px; background-color: ##{colour}; color: white;")
         elsif label == "recommend-accept"
           concat content_tag(:span, label, style: "padding: 3px; margin-right: 3px; border-radius: 2px; background-color: ##{colour}; color: #000000;")
+        elsif label == "query-scope"
```
arfon, line 40: This change means that papers labeled with `query-scope` on the `joss-reviews` repository will show the label in the dashboard.

lib/tasks/editorials.rake
```diff
@@ -21,4 +21,18 @@ namespace :editorials do
       end
     end
   end
+
+  desc "Send twice-weekly emails with possible submission rejections"
+  # Send on Monday and Thursday
+  task send_query_scope_email: :environment do
+    if Time.now.monday? || Time.now.thursday?
+      reviews_repo = Rails.application.settings["reviews"]
+      review_issues = ReviewIssue.download_review_issues(reviews_repo, 'query-scope')
```
arfon, line 30: Again, `query-scope` is the label we're including in the summary emails.

spec/models/vote_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe Vote, type: :model do
+  let(:vote) { create(:vote) }
```
arfon, line 4: #TODO write some tests...

arfon:
/ cc @openjournals/joss-eics @openjournals/joss-editors

danielskatz:
Do we want a whedon command to turn on the query-scope label?  Maybe not...

danielskatz:
otherwise, I like the idea of this

kthyng:
I like this a lot! Small thing: could the scope emojis only show up in the dashboard if the label has been activated for that submissions? Might make it cleaner and easier to see which are being discussed. Also, would there be a place that aggregates all of the submissions with this label (other than the email)?

eloisabentivegna:
Brilliant feature!

terrytangyuan:
Two questions 
1. How often do we receive emails like this from `noreply@joss`?
1. Maybe I missed it somewhere but where can we make comments instead of or along with voting?

arfon:
> Do we want a whedon command to turn on the query-scope label? Maybe not...

Something like `@whedon flag scope` or something like that? This might be a good way to ensure the correct label is applied, plus one could imagine using this in combination with asking the author to say more about the software on the `pre-review` issue.

arfon:
> How often do we receive emails like this from noreply@joss?

@terrytangyuan - the idea would be that these summary emails would be sent twice per week. Mondays and Thursdays and would only include those papers that have been flagged as possibly out of scope.

> Maybe I missed it somewhere but where can we make comments instead of or along with voting given that the email we receive cannot be replied?

Yes! Just zooming in a little on one of the screenshots from above, the idea would be that you can include a short (or long) comment with your vote in the text box:

Screen Shot 2020-07-01 at 3 53 27 PM

arfon:
> I like this a lot! Small thing: could the scope emojis only show up in the dashboard if the label has been activated for that submissions? Might make it cleaner and easier to see which are being discussed. Also, would there be a place that aggregates all of the submissions with this label (other than the email)?

Yes, that's a good idea. Perhaps we should just have a dedicated tab on the dashboard for these submissions?

labarba:
> the idea would be that these summary emails would be sent twice per week. Mondays and Thursdays and would only include those papers that have been flagged as possibly out of scope.

I _really_ think once per week is fine. No one, NO ONE, reacts to submissions at such a fast clip as we do. We are creating a monster.

jgostick:
It seems to me, in my brief 2 weeks under the deluge of these email conversations, that both **out of scope** and **lack of size** are issues that need input. In principle it might be possible to answer the size question quantitatively (like comparing the LOC to other projects in the same language), but this is probably most easily done just by asking someone familiar with the language.  In other words, should this feature do double duty for scope and size?

gkthiruvathukal:
@jgostick In a significant number of cases, I would think "lack of size" is also "lack of scope". Most of these end up being minor utilities and, therefore, not in scope.

This is a welcome addition, @arfon!

lib/tasks/editorials.rake
```diff
@@ -23,9 +23,9 @@ namespace :editorials do
   end
 
   desc "Send twice-weekly emails with possible submission rejections"
-  # Send on Monday and Thursday
+  # Send on Wednesday
```
danielskatz, line 26: Let's talk about this tomorrow - I would like to not only do this once a week, just because of the backlog that could build up and be carried over from one AEiC to another.  At least, we should think about the schedule of this with respect to AEiC rotation

danielskatz:
Once we do merge this in, we should check for open issues such as https://github.com/openjournals/joss/issues/709 that this PR might enable us to close

lib/tasks/editorials.rake
arfon: Sorry, we forgot to talk about this today.

My rationale for picking Wednesday was that this would be a couple of days after the start of an EiC rotation so potentially the incoming EiC would have had a chance to review the new submissions and flag any that potentially might be out of scope.

One alternative might be to send this on Sunday at the _end_ of a rotation?

kthyng: Wednesday seems ok to start. Maybe we should try through a full AEIC cycle and see what people think of the timing?

danielskatz: Right now, I feel like I own these decisions for submissions that come in while I am AEiC, and assuming I send out questions shortly after questionable submissions, I only can have 2 days after the end of my rotation that I need to pay attention.  If this goes out on Wednesdays, does that then mean that after my rotation is over, I have to pay attention for almost the full next week (W - F)?

PR #765, opened 2020-07-02
Show vote summary only when relevant

xuanxu:
This PR adds a change to the `scope-no-scope` branch to show the vote summary in the editors dashboard only for papers labeled with `query-scope`.


arfon:
:+1:

PR #774, opened 2020-07-13
adding COPYING as valid license file

danielskatz:
see https://github.com/openjournals/joss-reviews/issues/2400 for discussion that this will resolve


arfon:
:+1: thanks @danielskatz

PR #780, opened 2020-07-27
💎 Diamond OA

### LINKED ISSUE #775: add
It appears we are Diamond/platinum OA 

"Journals which publish open access without charging authors article processing charges are sometimes referred to as diamond or platinum OA. Since they do not charge either readers or authors directly, such publishers often require funding from external sources such as academic institutions, learned societies, philanthropists or government grants."  -- https://en.wikipedia.org/wiki/Open_access#Diamond/platinum_OA

Maybe we should use this term in our docs.

xuanxu:
This PR adds the OA variant to the `about` page.
Closes #775


arfon:
💎

PR #789, opened 2020-08-06
Make it a requirement that the paper has to be with the software

arfon:
Handling papers where the software isn't in a git repository _together with the paper_ is a huge pain for our editorial teams and breaks most (all?) of our automation. 

This PR makes what is currently a _strongly encouraged_ thing, compulsory.

/ cc @openjournals/joss-eics @openjournals/joss-editors

### DIFF
--- a/docs/submitting.md
+++ b/docs/submitting.md
@@ -10,7 +10,7 @@ But please read these instructions carefully for a streamlined submission.
 - The software should have an **obvious** research application.
 - You should be a major contributor to the software you are submitting.
 - Your paper should not focus on new research results accomplished with the software.
-- Your paper (`paper.md` and BibTeX files, plus any figures) must be hosted in a Git-based repository together with your software (although they may be in a different branch).
+- Your paper (`paper.md` and BibTeX files, plus any figures) must be hosted in a Git-based repository together with your software (although they may be in a short-lived branch which is never merged with the default).
 
 In addition, the software associated with your submission must:
 
--- a/docs/submitting.md
+++ b/docs/submitting.md
@@ -10,7 +10,7 @@ But please read these instructions carefully for a streamlined submission.
 - The software should have an **obvious** research application.
 - You should be a major contributor to the software you are submitting.
 - Your paper should not focus on new research results accomplished with the software.
-- Your paper (`paper.md` and BibTeX files, plus any figures) must be hosted in a Git-based repository. Placing these items together with your software (rather than in a separate repository) is **strongly encouraged**.
+- Your paper (`paper.md` and BibTeX files, plus any figures) must be hosted in a Git-based repository together with your software (although they may be in a different branch).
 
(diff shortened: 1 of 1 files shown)


danielskatz:
The only problem I see is that sometimes a paper is about software that is in more than one repository

arfon:
> The only problem I see is that sometimes a paper is about software that is in more than one repository

Yes, this is something that happens occasionally, I think we should allow submissions like this but waive the policy when this happens. (It's so unusual that I don't think we need to document this)

jni:
Hi everyone! Baby editor here so please disregard if my opinion is naive, but! I don't think that this requirement is particularly reasonable. I think for major software packages (NumPy, SciPy, or even scikit-image), the repo is a little bit "sacred" and non-code/non-doc things are generally discouraged from being in there. So personally I'd consider this a major disincentive to submitting to JOSS, and what's more, the more established the package, the bigger the disincentive, which might hold back JOSS's reputation.

I don't know enough about our automation to be properly informed here, but why couldn't we have a standard "link to software repo" format that lives in the paper repo (and defaults to `'.'`), and go from there?

xuanxu:
> the repo is a little bit "sacred" and non-code/non-doc things are generally discouraged from being in there. 

@jni As we allow submissions using any branch the easy solution for this is to use a temporary branch instead of master than can be deleted once the paper is published. If I remember correctly that's how some packages like [Sunpy](https://joss.theoj.org/papers/10.21105/joss.01832) did it.

arfon:
> @jni As we allow submissions using any branch the easy solution for this is to use a temporary branch instead of master than can be deleted once the paper is published. If I remember correctly that's how some packages like Sunpy did it.

That's correct. We have a fair number of submissions that come in these days (often more mature packages) where the JOSS paper never makes it into the commit history of the project, instead it lives in some kind of short-lived branch as @xuanxu mentioned (e.g. `joss-paper`). 

@jni I agree this could be a disincentive so perhaps we should make this clear in the submission docs?

jedbrown:
Is the protocol that any code/docs changes to the repository as a result of the review must be merged and in a tagged release (to be archived), but that the JOSS paper is just rebased along (or disjoint) and not part of the tagged archive? (I think this is fine, but we should clarify the expectation for this mode.)

danielskatz:
I believe this is correct.

arfon:
Yes that's correct.

jni:
Ah, ok, I didn't know about the branches, and even less that the branch could be short-lived. Thank you for clarifying this!

> perhaps we should make this clear in the submission docs?

Yes! =)

arfon:
OK, I've tweaked this a little more to make it clear the paper can live in a short-lived branch:

> Your paper (`paper.md` and BibTeX files, plus any figures) must be hosted in a Git-based repository together with your software (although they may be in a short-lived branch which is never merged with the default).

PR #812, opened 2020-10-03
Validate Git repository on submission

arfon:
Frequently, authors will submit papers with a Git address that can't be cloned. For example, the following are valid:

- https://github.com/arfon/fidgit
- https://github.com/arfon/fidgit.git
- git@github.com:arfon/fidgit.git

The following are not valid:

- https://github.com/arfon/fidgit/blob/master/paper/paper.md
- https://github.com/arfon/fidgit/tree/figure

The errors on submission usually arise because the author is trying to be specific about the branch or location of the `paper.md` file.

This pull request validates that the address can be `git cloned` (so Whedon will be able to handle it). The error message hopefully explains what's going wrong if an author submits something that won't work (suggestions welcome for improved language).

Screen Shot 2020-10-03 at 20 17 40

/ cc @openjournals/joss-eics

### DIFF
--- a/app/models/paper.rb
+++ b/app/models/paper.rb
@@ -1,3 +1,5 @@
+require 'open3'
+
 class Paper < ActiveRecord::Base
   searchkick index_name: "joss-production"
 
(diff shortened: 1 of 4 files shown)


xuanxu (commented):
🎉

Issue #819, opened 2020-10-18
Ask authors about parallel/related submissions


arfon:
When authors submit, we should ask them if any portion of the work is (or has been) being reviewed in another peer-reviewed publication venue.

Suggested language below:

> Has any portion of the submitted work (which includes code and documentation) been published or submitted, or is planned for submission, to another peer-reviewed publication venue?  If yes, please describe.

danielskatz:
I think this is reasonable, but we have to then figure out what rules we use when they say yes, where those rules are consistent across both AAS and other publishers

PR #821, opened 2020-10-20
Add co-publication guidelines to docs

kbarnhart:
Related to #819.

### DIFF
--- a/docs/review_criteria.md
+++ b/docs/review_criteria.md
@@ -46,6 +46,11 @@ These guidelines are not meant to be strictly prescriptive. Recently released so
 .. note:: The decision on scholarly effort is ultimately one made by JOSS editors. Reviewers are asked to flag submissions of questionable scope during the review process so that the editor can bring this to the attention of the JOSS editorial team.
 ```
 
+### Co-publication of science, methods, and software
+
+Sometimes authors prepare a JOSS publication alongside a contribution describing a science application, details of algorithm development, and/or methods assessment. In this circumstance, JOSS considers submissions for which the implementation of the software itself reflects a substantial scientific effort. This may be represented by the design of the software, the implementation of the algorithms, creation of tutorials, or any other aspect of the software. We ask that authors indicate whether related publications (published, in review, or nearing submission) exist as part of submitting to JOSS.
+
+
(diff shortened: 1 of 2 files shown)


danielskatz:
This seems to be in the review section of the docs, rather than the submission section - is this correct?

danielskatz:
Do we also want something here for reviewers that says what they should be looking for?  Is this part of one of our existing review criteria, perhaps significant scholarly content?

kbarnhart:
@danielskatz you are correct about the wrong place (should be fixed now). 🤦 

Not sure what I think about whether something should also be in the reviewer document. Will think on it.

arfon:
Thanks for doing this @kbarnhart.

> Do we also want something here for reviewers that says what they should be looking for? Is this part of one of our existing review criteria, perhaps significant scholarly content?

Our currently language is as follows:

> - [ ] Substantial scholarly effort: Does this submission meet the scope eligibility described in the JOSS guidelines

Perhaps we could change this to _Unique and substantial scholarly effort_:

> - [ ] Unique and substantial scholarly effort: Does this submission meet the scope eligibility described in the JOSS guidelines.

danielskatz:
@kbarnhart - I'm not sure how to change your suggested text, but I would prefer it if we could say something about what we would find problematic.

danielskatz:
@arfon - Your suggestions seems good to me

kbarnhart:
Agree that I like @arfon's suggestion. 

@danielskatz I think you should be able to edit the text via providing a review with suggestions. 

If that doesn't work, I just added you as a collaborator so you should now have push access to the branch.

danielskatz:
@kbarnhart - as I wrote "I'm not sure how to change your suggested text"...

kbarnhart:
@danielskatz, sorry if I'm being really dense today, do you mean that you don't know how you would like to change the content to add something about what would be problematic, or do you mean you are being blocked from modifying it. 

I had though the latter... 

I agree with you that adding that sort of element would be beneficial. Will think on it too.

danielskatz:
I meant I don't know what words to add

arfon (approved):

PR #822, opened 2020-10-21
Updated Karthik's profile pic

karthik:


arfon:
Thanks. I've actually just replaced your image with a reference to `https://github.com/karthik.png` - is this OK too? I'm gradually trying to remove the actual images of editors in the application and replace them with external URLs.

karthik:
Yes that's perfect. Easier for me to keep things updated here.

Issue #861, opened 2021-02-02
new high-level whedon command needed


danielskatz:
Some of AEiCs would like a new whedon command that combines some existing functionality:
- `generate pdf`
- `check references`
- check for statement of need
- check repo for license
this should also take a `from branch` option.  It might also be used within `accept`

kyleniemeyer:
Related, I've added an explicit mention of the statement of need to the review checklist in https://github.com/openjournals/joss/pull/862

kyleniemeyer:
For context to some of the above, right now some checks only occur at submission, and fail/are skipped if there are problems with the paper or repo initially. In particular, the check for an explicit Statement of Need section

xuanxu:
There is an existing command `check repository` that:
* detects license
* checks for statement of need section
* check languages
* analyzes code (general stats on authors and repo size)

This command is run automatically when a `[PRE-REVIEW]` issue is created, but can also be triggered with:
 `@whedon check repository`

kthyng:
@xuanxu Could this be triggered when the review issue is opened also? Sometimes we don't see that a Statement of Need is missing if it isn't caught by the pre-review automatic check repository when, for example, the paper can't be found and it isn't run properly. 

Alternatively, I could see checking for statement of need happening in the command to generate the pdf.

kyleniemeyer:
This is good to know! I think adding more info to [the docs](https://joss.readthedocs.io/en/latest/whedon.html#interacting-with-whedon) about what this does would help, I didn't realize it did all this.

I agree with @kthyng that having this automatically run at the start of the review issue is a good idea, just to confirm that major issues like this were resolved during the pre-review issue.

kyleniemeyer:
@xuanxu here's a question: how does `@whedon check repository` work when a paper is on a non-main branch, while the software is on the main branch? (For example, some people have the paper in a orphaned branch.)

danielskatz:
I guess we need a `from branch` option on check repository that can point to a paper and do the paper checks there, with the repo checks happening in the main branch

kthyng:
We could have people say which branch their paper is in if not in the master/main branch when they are submitting. This does come up pretty regularly.

xuanxu:
This feature is working with the new bot. Closing!

PR #868, opened 2021-02-05
Improving whedon command docs

kyleniemeyer:
The `@whedon query scope` command was missing from the docs


kyleniemeyer:
I also added some detail to the docs for the `whedon check repository` command, as per discussion in #861

xuanxu (approved):

PR #876, opened 2021-02-23
Active Editors List

xuanxu:
This PR adds information to the current editor list in the EiCs dashboard.
The new table includes this changes:
* `Show` action linked from the editor's name
* added link to GitHub page from the editor's avatar
* Editor login is now plain text easy to copy
* Description moved to appear only when hovering over the name of each editor
* Added a `Paper load` column with the current number of papers assigned to an editor, and linking to that editor's dashboard.
* Added availability info: status ([available/somewhat/none]) and comment
* Availability comment is a new editable field 
* Editors created in the last two months are marked as _new_  
* `New editor` button available before and after the editors table

---
**Current view:**
current_editors_list

---
**New view:**
screenshot_editors_new_page

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -409,8 +409,6 @@ TODO
     }
 }
 
-
-
 .pagination_helper {
     margin: 0.5em auto;
     display: block;
@@ -727,7 +725,7 @@ $btn-primary-border: darken($btn-primary-bg, 5%) !default;
     background-color: #FFB500 !important;
 }
 
-.paper-btn, .btn.paper-submit {
+.action-btn, .paper-btn, .btn.paper-submit {
     background-color: #2E294E;
     color: #fff;
     font-weight: bold;
@@ -744,6 +742,10 @@ $btn-primary-border: darken($btn-primary-bg, 5%) !default;
         color: #fff;
     }
 
+}
+.action-btn{
+ float: inline-end;
+ margin-bottom: 15px;
 }
 
 .btn.orcid {
--- a/app/assets/stylesheets/home.scss
+++ b/app/assets/stylesheets/home.scss
@@ -72,7 +72,7 @@ table.dashboard-table{
   thead {
     tr {
       th {
-        padding: 1em 0em 1em 1em;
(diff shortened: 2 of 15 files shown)


arfon:
👍 thanks for working on this. A couple of suggestions:

- I think the number of papers assigned should be sortable (see the `/dashboard` view for an example of how we're doing this in JavaScript already).
- I wonder if we can reduce the screen space give to the status strings. Perhaps we could swap these out for some kind of icon (🟢 / 🟠 / 🔴 ) and then have the comment about availability appear as some kind of mouseover?
- We can probably get rid of the 'type' column. I don't think we need this in the list view.

xuanxu:
Yes, I agree on everything, I will add the changes to this same PR

xuanxu:
@arfon ready!
* Icons to gain space 
* sortable columns 
* removed type info 
* availability shows an (*) if there is a comment and appears on mouseover:

corner

danielskatz:
We need to keep the editor's description visible, in my opinion - I use this to search on to find appropriate editors

danielskatz:
and could we make "show" for each editor a popup rather than an in-place new page?

kthyng:
If we are going to pursue having editor-selected max number of assignments, it might be more useful to list the number of submissions an editor has relative to their max (e.g. 2/4 or 1/1) and then do the red/yellow/green according to that rather than a static availability note. red is then if an editor has 4/4 or 1/1, yellow is if they have 3/4 and green is 0/4. Or maybe with this approach we only need two colors: 1 for more assignments and 1 for no more assignments.

kthyng:
I'm also still super interested in having at least some general categories or tags for area of expertise for editors (tags were mentioned in the call today). I think a descriptive list of topic areas is necessary to keep because people can do such varied and specific things, but another column with specific tags/categories so that as managing editor we can realize a submission is machine learning, then click on the ML editors and see who is green. That would be so amazing and save a ton of time!! We've discussed what categories to use before but could finalize a rough draft of which would be best to use.

arfon:
:+1: let's merge this and get it live and iterate further in future PRs.

PR #892, opened 2021-03-17
Editor profile and new availability calculation

xuanxu:
This PR adds an editor profile, so each editor can edit some of their data, including a max number of assignments, that is used to compute their availability.

Availability for each editor is based on the difference between max_assignments and the number of currently assigned (not paused) papers:
* if <= 0 the availability is none
* if 1 then availability is somewhat
* if >=2 then the editor is considered available

The PR includes a migration that will update all editors' availability to the new system. Default `max_assignments` value is 2.

new_profile

editor_profile


db/migrate/20210317094834_change_availability_to_max_assignments.rb
```diff
@@ -0,0 +1,23 @@
...
+    if Editor.first.respond_to?(:max_assignments) && Editor.first.respond_to?(:availability)
+      map_by_current_availability = { "available" => 2,
+                                      "somewhat" => 1,
+                                      "none" => 0,
+                                      "" => 2,
+                                      nil => 2 }
+      assigned_by_editor = Paper.unscoped.in_progress.group(:editor_id).count
+      Editor.emeritus.update_all(max_assignments: 0)
+      Editor.active.each do |editor|
+        editor.max_assignments = assigned_by_editor[editor.id].to_i + map_by_current_availability[editor.availability]
```
arfon, line 18: I'm not sure this is quite right. I'm currently editing 11 paper, and my max availability is marked as '13' but it shouldn't really be 11 + another 2.

Ideally I think we should assume that the goal of JOSS is to have 2-3 papers assigned to an editor at a time unless we explicitly hear from an editor that they are happy to handle more than this.

arfon (commented):
Thanks for opening this PR @xuanxu. I'm not completely convinced this is the right approach here. Specifically I'm wondering if it makes sense to remove the `availability` status when adding the `max_assignments` field.

As I understand the code, we're dropping the `availability` field in favor of determining this state by subtracting the number of assigned papers from the `max_assignments` value.

This makes sense in theory, but I worry that `max_assignments - assigned_papers` is not quite the same as `availability`. For example:

- I might desire to have no more than two papers assigned at a time, but I'm OK to be pushed a little beyond this at times because I'm _available_ to edit right now. 
- Conversely, I might be completely locked down with COVID challenges, have two papers already assigned to me, and _really really_ not be able to take on more papers right now. 

With this UI change, both of these are calculated and shown in the UI as the same editor status (red/yellow/green) but I don't think they're the same for a managing editor trying to figure out who can take another paper.

In summary, I think:

- `max_assignments - assigned_papers` is 'capacity' or how over-committed an editor is.
- `availability` is telling us something about the current situation of the editor and is useful to making decisions.

I'd be interested in what the other EiCs feel about this and how they interpret this information.

/ cc @openjournals/joss-eics

xuanxu:
> As I understand the code, we're dropping the `availability` field in favor of determining this state by subtracting the number of assigned papers from the `max_assignments` value.

Correct

> With this UI change, both of these are calculated and shown in the UI as the same editor status (red/yellow/green) but I don't think they're the same for a managing editor trying to figure out who can take another paper.

Yes, I had doubts too, but at the same time I find confusing the posibility of keeping both `availability` and `max_assignments` and have them giving contradictory information. I have let in place the `availability_comment` as a wildcard to provide extra information. 

So we'd be moving from:
`availability` + `availability_comment` only editable by EiCs
to
`max_assignments` + `availability_comment` auto assigned by each editor.

For this to work well, is responsability of the editors to update the `max_number` everytime their situation changes. In your two examples, the first could be solved with a `availability_comment` of "I'm OK to be pushed a little beyond this at times" and in the second case the editor can change the `max_assignments` to 0 for a while, and put it back to 2 or 3 once they have capacity again.

db/migrate/20210317094834_change_availability_to_max_assignments.rb
xuanxu: Yeah, it's a basic initial migration based on the current number of assignments and availability status. Here your max_assignments will be marked as 13 because you are currently editing 11 and available, so we assume you can still take 2 more. In the UI that will be something like: "Editing 11, Max 13"

kthyng:
I'm basing this comment on the comments above so I hope I am understanding it.

I think it is confusing to have two variables and would instead prefer to have only `max_assignments`. For the two scenarios this could look like:

> * I might desire to have no more than two papers assigned at a time, but I'm OK to be pushed a little beyond this at times because I'm _available_ to edit right now.

This should just be contained by `max_assignments` instead. If I am ok with 3 papers I should have `max_assignments=3` instead of it being sort of secret but ok if I am asked. 

> * Conversely, I might be completely locked down with COVID challenges, have two papers already assigned to me, and _really really_ not be able to take on more papers right now.

In this case, `max_assignments=0` so `max_assignments` - 2 = a negative number which means for sure you can't assign more submissions and might need to take on some of those already assigned (they could request that). 

Probably would help to have editors pinged once a month to re-choose their max number of submissions so low numbers don't just propagate forever.

PR #899, opened 2021-03-29
System specs

xuanxu:
This PR include the configuration for system specs


xuanxu (commented):
This removes the config code for specs from the "Editor profile" PR

PR #917, opened 2021-05-06
Valid Atom feed

### LINKED ISSUE #910: Atom feed has validation errors
I checked the [published papers Atom feed](https://joss.theoj.org/papers/published.atom) and [it contains some validation errors](https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Fjoss.theoj.org%2Fpapers%2Fpublished.atom), mainly coming from extraneous elements (`<state>`, `<software:version>`, `<submitted:at>`) and missing elements (`<author>`, since `<authors>` is used instead).

I noticed this because I was trying to use a RSS Reader and it failed (more specifically, the official Slack RSS integration<sup>1</sup>).

I am not an expert in Atom or XML but I suppose there is a way to extend the feed to add metadata in a way that still makes the feed valid?

<sub>1 It can also be the case that the official Slack RSS integration is not following the standard correctly with respect to foreign markup at the moment - but the failed validation still stands.</sub>

xuanxu:
This PR refactors the feed format to make it [valid Atom](https://validator.w3.org/feed/#validate_by_input).
- An general `author` element added to the feed 
- All entry's custom tags are nested under a generic `content` XML tag.

This should fix #910

### DIFF
--- a/app/views/papers/index.atom.builder
+++ b/app/views/papers/index.atom.builder
@@ -7,37 +7,43 @@ atom_feed do |feed|
   feed.link(rel: 'last', type: "application/atom+xml", href: url_for(params: url_params, format: 'atom', page: @papers.total_pages, only_path: false))
   feed.title(Rails.application.settings["name"])
   feed.updated(@papers[0].created_at) if @papers.length > 0
+  feed.author do |author|
+    author.name(Rails.application.settings["name"])
+    author.uri(Rails.application.settings["url"])
+  end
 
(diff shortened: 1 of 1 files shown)


arfon:
✨ https://validator.w3.org/feed/check.cgi?url=https%3A//joss.theoj.org/papers/published.atom

astrojuanlu:
Oh, didn't expect such a quick response! Thanks folks 🙌🏽

PR #944, opened 2021-06-16
Onboarding

xuanxu:


lib/repository.rb
```diff
@@ -17,4 +17,24 @@ def self.editors
...
+      nil
+    end
+
+    if invitee_id.present?
+      org = Rails.application.settings["reviews"].split("/").first
+      url = "https://api.github.com/orgs/#{org}/invitations"
+      headers = {"Authorization" => "token #{ENV['GH_TOKEN']}", "Content-Type" => "application/json", "Accept" => "application/vnd.github.v3+json"}
+      parameters = {invitee_id: invitee_id, team_ids: [config[nwo]["editor_team_id"]]}
+
+      response = Faraday.post(url, parameters.to_json, headers)
```
arfon, line 34: Only issue with this in production is that Whedon isn't an org admin so this line of code currently fails.

xuanxu: Right! I have to document that, the GH_TOKEN needs admin privileges

xuanxu: Actualy I think Whedon is an org admin, it's already sending invitations and adding collaborators to the organization. If its token includes the admin:org scope his should not fail.

xuanxu (commented):
Added fields to the new editor form: 

neweditor

lib/repository.rb
arfon: :+1:

PR #955, opened 2021-07-06
Remove payload parsing code

xuanxu:
This PR removes the code for webhook payload parsing and delegates that task to the Issue gem.


xuanxu (commented):
Recommended: deploy this branch and manually test it before merging

PR #958, opened 2021-07-06
Making it clear COI applies to editors too

arfon:
Making it clear that COI policies apply to editors too.

/ cc @emdupre who raised this.


docs/reviewer_guidelines.md
```diff
@@ -19,8 +19,8 @@ You can include in your review links to any new issues that you the reviewer bel
...
+If you have a conflict of interest with a submission, you should disclose the specific reason to the submissions' editor (or to an EiC if you are an editor). This may lead to you not being able to review or edit the submission, but some conflicts may be recorded and then waived, and if you think you are able to make an impartial assessment of the work, you should request that the conflict be waived. For example, if you and a submitter were two of 2000 authors of a high energy physics paper but did not actually collaborate. Or if you and a submitter worked together 6 years ago, but due to delays in the publishing industry, a paper from that collaboration with both of you as authors was published 2 year ago. Or if you and a submitter are both employed by the same very large organization but in different units without any knowledge of each other.
```
danielskatz, line 24:
```suggestion
If you have a conflict of interest with a submission, you should disclose the specific reason to the submissions' editor (or to an EiC if you are an editor, or to the other EiCs if you are an EiC). This may lead to you not being able to review or edit the submission, but some conflicts may be recorded and then waived, and if you think you are able to make an impartial assessment of the work, you should request that the conflict be waived. For example, if you and a submitter were two of 2000 authors of a high energy physics paper but did not actually collaborate. Or if you and a submitter worked together 6 years ago, but due to delays in the publishing industry, a paper from that collaboration with both of you as authors was published 2 year ago. Or if you and a submitter are both employed by the same very large organization but in different units without any knowledge of each other.
```

danielskatz:
other than my suggested change for even more detail and clarity, this looks good to me

emdupre:
Agreed with @danielskatz's suggested addition, but otherwise this LGTM to me and clarifies any confusion ! Thank you both !

PR #964, opened 2021-07-19
review_criteria: remove specific call-out of Travis-CI

jedbrown:
There are many continuous integration providers, hosted and not.
Travis-CI changed their open source policy and many packages moved away.
GitHub Actions is the most convenient for projects on GitHub first
considering CI, but docs are accessible enough we don't need to plug
them specifically.

### DIFF
--- a/docs/review_criteria.md
+++ b/docs/review_criteria.md
@@ -94,7 +94,7 @@ Reviewers are expected to install the software they are reviewing and to verify 
 
 Authors are strongly encouraged to include an automated test suite covering the core functionality of their software.
 
-> **Good:** An automated test suite hooked up to an external service such as Travis-CI or similar<br />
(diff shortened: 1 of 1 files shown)


danielskatz:
In other places in docs and messaging, we try to list some options.  Do you think it's better not to change to "GitHub Actions, Circle CI, or similar"?

jedbrown:
I'm not sure because we're putting a thumb on the scale for commercial interests (and raising barriers to entry). A search for `GitHub continuous integration` returns GitHub Actions, which is promoted in various places on GitHub already. It's probably the best option for someone who would need to search at this point while reading the docs. We could list GitHub Actions, GitLab-CI, Circle-CI, Azure DevOps, Appveyor, for example. I have no strong feelings either way.

arfon:
I added some options :smile:. Thanks @jedbrown and @danielskatz .

PR #972, opened 2021-08-10
Proper caption to notification settings screenshot

martinfleis:
While reading through the editorial guide I've noticed a trailing word (see the screenshot) which is actually the figure caption. 

I've changed it to a meaningful phrase.

The captions are normally not shown when rendering markdown (like on GitHub) but sphinx seems to treat them differently in this case.

Screenshot 2021-08-10 at 17 46 07

My linter also removed trailing white spaces in the file, sorry for that noise.


arfon:
✨ thanks @martinfleis!

Issue #680, opened 2020-02-12
can reviewers ask for best practices in the installation of packages?


pdebuyl:
My question stems from the recommendation, for Python packages, to use

    pip install package

instead of

    pip install --user package

The first version will error except if the user uses a virtualenvironment because of permissions. Googling the issue will pop up the bad advice to just use sudo. This can damage system installations of Python.

May reviewers ask to change the instruction to either use the `--user` flag or to suggest using virtualenvironments? 

The issue of best practices was already discussed in #469 

I realize that this is a Python specific question, but the general principle should be useful for other tools as well.

pdebuyl:
From https://journals.plos.org/plosbiology/article?id=10.1371/journal.pbio.1001745

The paper "Best Practices for Scientific Computing"

> Box 1. Summary of Best Practices
> Write programs for people, not computers.
> A program should not require its readers to hold more than a handful of facts in memory at once.
> Make names consistent, distinctive, and meaningful.
> Make code style and formatting consistent.
> Let the computer do the work.
> Make the computer repeat tasks.
> Save recent commands in a file for re-use.
> Use a build tool to automate workflows.
> Make incremental changes.
> Work in small steps with frequent feedback and course correction.
> Use a version control system.
> Put everything that has been created manually in version control.
> Don't repeat yourself (or others).
> Every piece of data must have a single authoritative representation in the system.
> Modularize code rather than copying and pasting.
> Re-use code instead of rewriting it.
> Plan for mistakes.
> Add assertions to programs to check their operation.
> Use an off-the-shelf unit testing library.
> Turn bugs into test cases.
> Use a symbolic debugger.
> Optimize software only after it works correctly.
> Use a profiler to identify bottlenecks.
> Write code in the highest-level language possible.
> Document design and purpose, not mechanics.
> Document interfaces and reasons, not implementations.
> Refactor code in preference to explaining how it works.
> Embed the documentation for a piece of software in that software.
> Collaborate.
> Use pre-merge code reviews.
> Use pair programming when bringing someone new up to speed and when tackling particularly tricky problems.
> Use an issue tracking tool.
> 

or https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1005510

the paper "Good enough practices in scientific computing"

> Box 1. Summary of practices
> Data management
> Save the raw data.
> Ensure that raw data are backed up in more than one location.
> Create the data you wish to see in the world.
> Create analysis-friendly data.
> Record all the steps used to process data.
[...]

Issue #469, opened 2018-12-01
strengthen FOSS best practices as criterion for joss submissions


maxheld83:
I recently an [R package](https://github.com/openjournals/joss-reviews/issues/911) and came away from the process a) in love the vision of joss, and b) a little confused about the importance of FOSS best practices in joss submissions.

The software I reviewed was, I thought, too needlessly specialised, monolithic in its design, and generally lacking in software dev / FOSS best practices **to be of long-term, maintainable use to the community**.
(The software essentially worked only on/with the *included* 4 datasets, and perhaps was more of an interactive vignette overall, than a fully-fledged package).
I worry that these kinds of packages really fragment the (in this case) R ecosystem, and degrade the user experience. 
They also underline the stereotype that scientific software is usually pretty poorly written software.

I understand that scientists have  a *lot* of other obligations, and little training in software development (I am in that same situation).

However, if **we really want to strengthen open source contributions as proper *scientific* contributions**, I think we should really focus on **high quality packages, with good maintainability, modularity, a clear focus** and all the other evolved FOSS best practices.

Otherwise, in the extreme, we're risking a situation where people are now gaming "software contributions" with high-frequency, low-quality material, much like they have been gaming "paper contributions", to the detriment of careful scientific software developers and the overall open science enterprise.

Obviously, this is a thin needle to thread: standardise demands too much, and you end up with a calcified system (for example, I wouldn't want to only accept "tidyverse" packages for R, that'd be crazy).

For the review in question, @arfon's decision seems right to me – it'd be unfair to spring such demands on the submitting authors @davidnsousa and @joaordaniel at this point.

And I imagine that you fine people at foss have had plenty of discussions on this issue.

So I just wanted to report my experience here.

In my view, open science might be better off if we give credit for high quality research software, and be very clear about what makes for such software in advance.
**Quality before quantity**.

arfon:
Thanks for posting this @maxheld83 - it's always good to hear the experience of reviewers, especially those new to JOSS.

I basically agree with everything you say here, but want to thank you for your support with the decision on https://github.com/openjournals/joss-reviews/issues/911. 

> However, if we really want to strengthen open source contributions as proper scientific contributions, I think we should really focus on high quality packages, with good maintainability, modularity, a clear focus and all the other evolved FOSS best practices.

[...]

PR #976, opened 2021-08-21
Adding packaging requirement to submission docs

arfon:
Many submissions to JOSS require significant work to improve their packaging. We don't really want reviewers to have to point this out to authors, instead we should catch this at the pre-review stage and hold back from review until the software is appropriately packaged.

Examples of guidelines we could point authors to include:

- https://packaging.python.org
- https://r-pkgs.org/index.html
- https://www.digitalocean.com/community/tutorials/how-to-package-and-distribute-ruby-applications-as-a-gem-using-rubygems

Related to https://github.com/openjournals/joss/issues/680 and https://github.com/openjournals/joss/issues/469

### DIFF
--- a/docs/review_criteria.md
+++ b/docs/review_criteria.md
@@ -56,7 +56,7 @@ The authors should clearly state what problems the software is designed to solve
 
 #### Installation instructions
 
-Software depdendencies should be clearly documented and their installation handled by an automated proceedure. Where possible, software installation should be managed with a package manager. However, this criterion depends upon the maturity and availability of software packaging and distribution in the programming language being used. For example, Python packages should be `pip install`able, and have adopted [packaging conventions](https://packaging.python.org), while Fortran submissions with a Makefile may be sufficient.
+Software dependencies should be clearly documented and their installation handled by an automated proceedure. Where possible, software installation should be managed with a package manager. However, this criterion depends upon the maturity and availability of software packaging and distribution in the programming language being used. For example, Python packages should be `pip install`able, and have adopted [packaging conventions](https://packaging.python.org), while Fortran submissions with a Makefile may be sufficient.
 
 > **Good:** The software is simple to install, and follows established distribution and dependency management approaches for the language being used<br />
 > **OK:** A list of dependencies to install, together with some kind of script to handle their installation (e.g., a Makefile<br />
--- a/docs/review_criteria.md
+++ b/docs/review_criteria.md
@@ -56,11 +56,11 @@ The authors should clearly state what problems the software is designed to solve
 
 #### Installation instructions
 
-There should be a clearly-stated list of dependencies. Ideally these should be handled with an automated package management solution.
+Software depdendencies should be clearly documented and their installation handled by an automated proceedure. Where possible software installation should be managed with a package manager, however this criterion depends upon the maturity and availability of software packaging and distribution in the programming language being used. For example, Python packages should be `pip install`able, and have adopted [packaging conventions](https://packaging.python.org), Fortran submissions with a Makefile may be sufficient.
 
-> **Good:** A package management file such as a Gemfile or package.json or equivalent<br />
-> **OK:** A list of dependencies to install<br />
-> **Bad (not acceptable):** Reliance on other software not listed by the authors
+> **Good:** The software is simple to install, and follows established distribution and dependency management approaches for the language being used<br />
+> **OK:** A list of dependencies to install, together with some kind of script to handle their installation (e.g., a Makefile<br />
+> **Bad (not acceptable):** Dependencies are unclear, and/or installation process lacks automation
 
 #### Example usage
 
--- a/docs/review_criteria.md
+++ b/docs/review_criteria.md
@@ -56,7 +56,7 @@ The authors should clearly state what problems the software is designed to solve
 
 #### Installation instructions
 
-Software depdendencies should be clearly documented and their installation handled by an automated proceedure. Where possible software installation should be managed with a package manager, however this criterion depends upon the maturity and availability of software packaging and distribution in the programming language being used. For example, Python packages should be `pip install`able, and have adopted [packaging conventions](https://packaging.python.org), Fortran submissions with a Makefile may be sufficient.
+Software depdendencies should be clearly documented and their installation handled by an automated proceedure. Where possible, software installation should be managed with a package manager. However, this criterion depends upon the maturity and availability of software packaging and distribution in the programming language being used. For example, Python packages should be `pip install`able, and have adopted [packaging conventions](https://packaging.python.org), while Fortran submissions with a Makefile may be sufficient.
 
 > **Good:** The software is simple to install, and follows established distribution and dependency management approaches for the language being used<br />
 > **OK:** A list of dependencies to install, together with some kind of script to handle their installation (e.g., a Makefile<br />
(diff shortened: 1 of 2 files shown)


danielskatz:
Should we provide this guidance in our instructions? I think it would be better for submitters if we did so.

jedbrown:
I'm in favor of this, and think we need to provide pointers and some caveats so people who don't know how to do this (or for whom this would be especially arduous) can still get some advice. Most "modern" languages have packaging systems, but for example, packaging Python HPC-ish software (anything with native language extension modules) in PyPI is quite arduous, to the extent that well-funded industry projects sometime skip it or need a level of effort that is unattainable by most small projects.

I also think there's a philosophy issue that can be delicate and a potential source of dispute with reviewers (conda, nix, PyPI/CRAN/crates.io, deb, rpm, brew, spack). Perhaps rather than giving a rubric or asking reviewers if it's packaged in the system that they use, just say that it should be packaged in _some_ relevant system that can be tested (hopefully at least one reviewer can run `$tool install` in docker or WSL if the packager doesn't play well with their daily environment).

Going through the process of packaging is a great learning experience if not endlessly hung up on poorly-documented minutia.

kthyng:
I recently learned a lot about packaging myself by leaning heavily on a cookie cutter template, and I found it incredibly useful. Should we recommend this sort of thing too? Here is the base version: https://github.com/cookiecutter/cookiecutter, but of course there are various more specific versions we could give as examples too. I used the "sciencey" one here: https://github.com/jbusecke/cookiecutter-science-project.

danielskatz:
I was thinking about cookie cutters too, but wasn't sure of good examples for packaging code as opposed to doing science more reproducibly

dfm:
This might be Too Much™, but another Python-specific link (it's not really astronomy-specific, despite the name) that I usually share with collaborators is: https://packaging-guide.openastronomy.org/en/latest/index.html

pdebuyl:
In the review criteria, there is:

> Ideally these should be handled with an automated package management solution.

that falls under the "installation instructions" checklist item.

I am "+1" anyway to add the suggested wording to the "submitting" page.

I also agree that we should not put in our instructions "accepted packaging tools". We could add something in the criteria about the effort to install the package. For instance, for a Fortran code a src directory with an example makefile *can* be an acceptable solution. For Python, something along `pip install .` should be possible. I am being too specific here for the instructions, to emphasize the necessity to remain based on principles and on the evaluation by language/field experts.

arfon:
:wave: all, would love your thoughts on these changes. I have attempted to address your feedback and added some more words to the review criteria.

docs/review_criteria.md
```diff
@@ -56,11 +56,11 @@ The authors should clearly state what problems the software is designed to solve
 
 #### Installation instructions
 
-There should be a clearly-stated list of dependencies. Ideally these should be handled with an automated package management solution.
+Software depdendencies should be clearly documented and their installation handled by an automated proceedure. Where possible software installation should be managed with a package manager, however this criterion depends upon the maturity and availability of software packaging and distribution in the programming language being used. For example, Python packages should be `pip install`able, and have adopted [packaging conventions](https://packaging.python.org), Fortran submissions with a Makefile may be sufficient.
```
danielskatz, line 59:
```suggestion
Software depdendencies should be clearly documented and their installation handled by an automated proceedure. Where possible, software installation should be managed with a package manager. However, this criterion depends upon the maturity and availability of software packaging and distribution in the programming language being used. For example, Python packages should be `pip install`able, and have adopted [packaging conventions](https://packaging.python.org), while Fortran submissions with a Makefile may be sufficient.
```

danielskatz:
I suggested some minor changes to wording, but generally think this looks good.

docs/review_criteria.md
fabian-s: Software depdendencies --> Software dependencies

arfon: > Software depdendencies --> Software dependencies

Good catch. Thanks!

docs/review_criteria.md
```diff
@@ -56,11 +56,11 @@ The authors should clearly state what problems the software is designed to solve
 
 #### Installation instructions
 
-There should be a clearly-stated list of dependencies. Ideally these should be handled with an automated package management solution.
+Software dependencies should be clearly documented and their installation handled by an automated proceedure. Where possible, software installation should be managed with a package manager. However, this criterion depends upon the maturity and availability of software packaging and distribution in the programming language being used. For example, Python packages should be `pip install`able, and have adopted [packaging conventions](https://packaging.python.org), while Fortran submissions with a Makefile may be sufficient.
```
faroit, line 59: for python: does this mean we enforce pypi packaging and not allow conda-forge?

docs/submitting.md
```diff
@@ -36,7 +36,7 @@ As a rule of thumb, JOSS' minimum allowable contribution should represent **not
 - Whether the software has already been cited in academic papers.
 - Whether the software is sufficiently useful that it is _likely to be cited_ by your peer group.
 
-In addition, JOSS requires that software should be feature-complete (i.e. no half-baked solutions) and designed for maintainable extension (not one-off modifications of existing tools). "Minor utility" packages, including "thin" API clients, and single-function packages are not acceptable.
+In addition, JOSS requires that software should be feature-complete (i.e., no half-baked solutions), packaged appropriately according to common community standards for the programming language being used (e.g., [Python](https://packaging.python.org), [R](https://r-pkgs.org/index.html)), and designed for maintainable extension (not one-off modifications of existing tools). "Minor utility" packages, including "thin" API clients, and single-function packages are not acceptable.
```
faroit, line 39: Not sure if this is relevant here but in the past, I saw problems communicating proper package renaming (e.g. when the package name wasn't appropriate for JOSS). Not sure if this is reflected in the community guides. e.g. see https://www.python.org/dev/peps/pep-0423/#how-to-rename-a-project

docs/review_criteria.md
kthyng: I would lean toward listed both as examples ("Python packages should be `pip install`able and/or `conda install`able with the conda-forge channel...") to be clear but certainly being installable with conda-forge should be ok. 

Is it ok for a package to say "clone this from github and install locally with `pip install -e .`?

faroit: > Is it ok for a package to say "clone this from github and install locally with pip install -e .?

@kthyng I can't think of a reason why a dev install would be a good default choice compared to hosting the package on pypi or conda-forge.

kthyng: @faroit I agree, but I think we may see this come through. I guess it is covered by what it says above, though (about needing to be pip installable).

kthyng: No I take it back, @faroit, @arfon I think we should specify that it can't be dev installable only since people might think it is pip installable since it uses "pip". This is another thing a possibly not-software-oriented person would have to learn to do though.

jedbrown: There's also a branding/namespace issue, in that if we publish a paper on a Python package and `pip install foo` works, the reader should be getting the package they're reading about. If we only check that local installs work, a different package (now or in the future) could take that name in PyPI. Of course we can't enforce this across all packaging systems, but one or two are usually pervasive within any given community.

This gets more complicated fast for software that is not single-language. I'd also note that sometimes software is packaged for binary distribution, but developer installs are really hard. That's enough for typical users, but unhealthy for projects that need users to wet their feet by contributing. Should our requirements say that software should be packaged (at least one relevant place) *and* that developer installs are easy to follow/play well with others? (For example, they shouldn't require a custom Docker image to build.)

arfon:
Merging these changes as-is. If there are further changes let's discuss in additional/follow-up PRs.

PR #996, opened 2021-09-30
Refactor checklists

xuanxu:
This PR removes all the code related with removing reviewer checklists from the Review body.

Ref: Should be deployed at the same time as openjournals/whedon-api#138


danielskatz:
I find percentage complete to be very useful - do we need to remove it?

arfon:
> I find percentage complete to be very useful - do we need to remove it?

Me too. Unfortunately the current implementation will break if we make this change 😞. So I guess the question is, would we rather keep the percentage complete information, or address the usability challenges of the repository invitation flow that reviewers have to go through.

danielskatz:
there's no way to work around the change and still have this?

xuanxu:
Closing as this PR is not needed anymore

PR #1013, opened 2021-11-19
Expire invitations to rejected papers

xuanxu:
When a paper is rejected all the pending invitations to edit should be expired.


arfon:
Nice!

PR #1028, opened 2021-12-20
Add warning to users about figure file in draft paper example

VLucet:
See https://github.com/openjournals/openjournals-draft-action/issues/3. I suggested to comment out the figure code in the draft example for users who would like to try the github action. The code should remain to illustrate how to insert figures.


danielskatz:
👋 @tarleb - Is this still a meaningful PR to resolve and merge, given the changes in JOSS and its tooling since this was created?

PR #1033, opened 2022-01-13
taking out step that GitHub does for you

danielskatz:


arfon:
Actually _Whedon/JOSS_ does this for you :-)

PR #1034, opened 2022-01-14
Draft: Statement of need: context of related work

jedbrown:
I find it frustrating when software and JOSS submissions don't mention related work, which sometimes overlaps heavily with the new capability. While JOSS papers don't have a scientific novelty standard, I think we should be explicit that authors are expected to cite alternatives, especially well-known alternatives, and explain what they're bringing to the table.

This is `Draft` because the text is reproduced in various other places (perhaps generated files) and I'd like to get consensus wording before propagating.

### DIFF
--- a/docs/review_checklist.md
+++ b/docs/review_checklist.md
@@ -40,7 +40,7 @@ Below is an example of the review checklist for the [Yellowbrick JOSS submission
 ### Software paper
 
 - **Summary:** Has a clear description of the high-level functionality and purpose of the software for a diverse, non-specialist audience been provided?
-- **A statement of need:** Does the paper have a section titled 'Statement of Need' that clearly states what problems the software is designed to solve and who the target audience is?
+- **A statement of need:** Does the paper have a section titled 'Statement of need' that clearly states what problems the software is designed to solve, who the target audience is, and its relation to other work?
 - **State of the field:** Do the authors describe how this software compares to other commonly-used packages?
 - **Quality of writing:** Is the paper well written (i.e., it does not require editing for structure, language, or writing quality)?
 - **References:** Is the list of references complete, and is everything cited appropriately that should be cited (e.g., papers, datasets, software)? Do references in the text use the proper [citation syntax]( https://rmarkdown.rstudio.com/authoring_bibliographies_and_citations.html#citation_syntax)?
--- a/docs/submitting.md
+++ b/docs/submitting.md
@@ -78,12 +78,12 @@ Your paper should include:
 
 - A list of the authors of the software and their affiliations, using the correct format (see the example below).
 - A summary describing the high-level functionality and purpose of the software for a diverse, *non-specialist audience*.
-- A *Statement of Need* section that clearly illustrates the research purpose of the software.
+- A *Statement of need* section that clearly illustrates the research purpose of the software and places it in the context of related work.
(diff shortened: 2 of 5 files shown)


danielskatz:
I prefer "prior work" to "prior works"

danielskatz:
I'm also unsure of "Statement of Need" vs "Statement of need"

pdebuyl:
We could also consider "related work" instead of "prior work". A JOSS paper might describe a software that predates the alternatives, hence there would be no other prior work despite there being alternatives.

docs/review_checklist.md
```diff
@@ -30,7 +30,7 @@ Below is an example of the review checklist for the [Yellowbrick JOSS submission
 
 ### Documentation
 
-- **A statement of need:** Do the authors clearly state what problems the software is designed to solve and who the target audience is?
+- **A statement of need:** Do the authors clearly state what problems the software is designed to solve, who the target audience is, and its relation to prior works?
```
jedbrown, line 33: @pdebuyl "relation to related work" seemed clumsy given the sentence structure, but I agree that "related work" is better language...

docs/submitting.md
```diff
@@ -78,12 +78,12 @@ Your paper should include:
 
 - A list of the authors of the software and their affiliations, using the correct format (see the example below).
 - A summary describing the high-level functionality and purpose of the software for a diverse, *non-specialist audience*.
-- A *Statement of Need* section that clearly illustrates the research purpose of the software.
+- A *Statement of need* section that clearly illustrates the research purpose of the software and places it in the context of related work.
```
jedbrown, line 81: @pdebuyl @danielskatz Is there agreement on this wording and it just needs to be massaged into sentence structure elsewhere?

danielskatz: It's ok with me

pdebuyl: yes, that is fine!

docs/review_checklist.md
jedbrown: Should we have this requirement (discussing related work) for both the project documentation and the JOSS paper, or would it be acceptable for it to only be in the JOSS paper?

My personal view is that it's necessary in the paper and nice to have in project documentation. However, it can be hard to maintain in project documentation because other packages are created and existing ones develop new features such that rationale can become stale. I think it's good for maintainers to stay informed about major capabilities of peers/competitors and end-to-end workflows, but perhaps it's too much to ask. I think it could be harmful for professional relationships if package X says something is a unique killer feature not available in package Y, and yet package Y does that thing in its next release but the page doesn't get updated for years. It could be resolved with a friendly email, but people don't always do that.

pdebuyl: In my opinion, it is ok to require it only in the article, as it is a snapshot of the software and of the landscape at the time of publication. Also, we don't have control on the future of the packages (and their doc) that we review.

danielskatz: I'll agree with it being more important (and time bound) in the paper

arfon: > and its relation to other work

I agree with @danielskatz that 'other' is the right thing here. Rationale we want people to mention other work (not by them) whereas 'prior' might suggest prior work by this team?

jedbrown: Thanks, I've updated this PR. I don't know what to do with this sentence fragment appearing various places in `spec/fixture/`. I understand these are tests with previously published papers so perhaps the wording should not be updated because it's comparing with the state of the old paper. Or perhaps it's meant to be a dry-run with the content of the old paper in the new environment, but the CI action passed :shrug:.

PR #1036, opened 2022-01-17
Removing @adrn's ORCID from docs

dfm:
He's currently getting assigned a lot of random JOSS papers because authors don't seem to update the ORCID in their submissions.


arfon:
Good idea. Sorry @adrn for all the noise here...

adrn:
All good!! Thanks for the quick updates :)

PR #1040, opened 2022-01-31
Add information on how to update the Zenodo record

diehlpk:
Recently, some author had a different title and list of authors for the Zenodo archive. They were not aware how to change that, and we had to provide these details. I added some instructions to the documentation.

### DIFF
--- a/docs/example_paper.md
+++ b/docs/example_paper.md
@@ -31,6 +31,7 @@ authors:
 affiliations:
  - name: Lyman Spitzer, Jr. Fellow, Princeton University, USA
    index: 1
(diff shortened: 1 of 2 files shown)


danielskatz (approved):
lgtm

emdupre:
Would it make sense to put each sentence on a new line for easy diff-ing in the future?

Also, it might make sense to directly link to the Zenodo guide in case they update the procedure: https://help.zenodo.org/#technical

and/or their note re: editing the title does not mint a new DOI : https://help.zenodo.org/#versioning

PR #1065, opened 2022-04-19
Link to paper source in docs

tarleb:
The bit.ly link points to an outdated version that's missing
important parts of a paper, e.g., the "statement of need".


xuanxu (approved):
:+1:

PR #1086, opened 2022-06-05
Adding initial tracks

arfon:
Adding initial track definitions based on conversation in earlier [Google doc](https://docs.google.com/document/d/12vowUKTfsPiPMg6Gny2g9YTLLISZiL0yRjCwjloeFDE/edit#heading=h.8qbx1hgq2slt).

This track list will be used to replace the 'suggested editor' on the JOSS submission form, instead, we will require authors to pick a subject area for their paper. The list of subjects is a controlled one (see `spec/fixtures/reference-tracks.yml`), with the JOSS track definitions derived from this (`lib/tracks.yml`)

Screenshot_2022-06-05_at_14_15_20

Note this needs review and input from initial JOSS track EiCs (@arfon, @Kevin-Mattheus-Moerman, @kyleniemeyer, @oliviaguest, @gkthiruvathukal, @kthyng, @danielskatz) to bootstrap the subject-JOSS track mapping.

**What I need your review on please**

Please take a look a the file `lib/tracks.yml` and review for track 'fit'. If you don't feel like something fits in your track then please suggest which track it should be in. 

Some other general thoughts:

- There are many subject areas here that we don't receive submissions in JOSS for today – don't worry about that. If we start to receive lots of medicine/healthcare submissions then we can make a new track!
- Please don't make additions, I think we should keep to this current list of subjects if we can.
- Some track areas have waaaay more subjects than others. I *think* this is OK, although I am a little concerned about the very limited number of subjects under _Data Science, Artificial Intelligence, and Machine Learning_ / cc @gkthiruvathukal


danielskatz:
In general, I think we should provide some clear guidance to authors about what to do if they find a problem with this, then to some extent, we can trial and error it out

kyleniemeyer:
@danielskatz Ocean Engineering is more about engineering things that go in the ocean, so I think it should be into Physics and Engineering.

I also think a bunch of topics currently in Earth Sciences and Ecology should also go in Physics and Engineering:
       - General Energy [Physical Sciences]
       - Energy (miscellaneous) [Physical Sciences]
       - Energy Engineering and Power Technology [Physical Sciences]
       - Fuel Technology [Physical Sciences]
       - Nuclear Energy and Engineering [Physical Sciences]
       - Renewable Energy, Sustainability and the Environment [Physical Sciences]
       - Pollution [Physical Sciences]
       - Waste Management and Disposal [Physical Sciences]

gkthiruvathukal:
Hi All, just getting to this now. Please stay tuned for some specific feedback about the my prospective track.

gkthiruvathukal:
For **AI and ML**, I would like to start by using the ACM scheme, https://cran.r-project.org/web/classifications/ACM.html:

ARTIFICIAL INTELLIGENCE [and MACHINE LEARNING]:

- General
- Applications and Expert Systems
- Automatic Programming
- Deduction and Theorem Proving
- Knowledge Representation Formalisms and Methods
- Programming Languages and Software
- Learning
- Natural Language Processing
- Problem Solving, Control Methods, and Search
- Robotics
- Vision and Scene Understanding (rename to "Computer Vision")
- Distributed Artificial Intelligence

Although a good number of these are more applied than others, this would be a great start as it includes most topics in modern AI/ML practice.

I'd also consider adding:

- Recommender Systems
- Low-Power/Energy Efficient Methods
- Agent-Based Systems
- Applications

As for **Data Science**, I would like to propose that this be kept as general as possible. Taxonomies for this are still emerging, but I think these possibilities come to mind:

- Statistical Libraries and Tools
- Modeling and Simulation [using statistical methods]
- Decision Analysis
- Linear Programming and Optimization

I hope this helps, @arfon.

arfon:
Hi @gkthiruvathukal – I really want us to leverage a controlled list here rather than building our own. Although I agree that the subject areas for AI/ML are pretty deficient. It seems like adding our own subject area categories is opening up a can of worms here...

gkthiruvathukal:
@arfon agreed. The first part of my list comes from ACM. I only suggested additions based on current practice.

On the other hand, I am comfortable with the major topics as headings (data science, AI/ML).

danielskatz:
I'm ok to try this with the tracks as now written, and making changes after a bit if needed.

However, I'm still concerned that this change will make us less responsive to authors, since track editors won't really have a way of taking a break, unless we consider backups/deputies in some way.

kyleniemeyer:
> > I'm ok to try this with the tracks as now written, and making changes after a bit if needed.
> 
> 👍 thanks.
> 
> > However, I'm still concerned that this change will make us less responsive to authors, since track editors won't really have a way of taking a break, unless we consider backups/deputies in some way.
> 
> Yes, I'm worried about this too. A few things I've thought about that could potentially mitigate this risk:
> 
> * Set up per-track Slack channels (and have editors aligned with those tracks be present in those channels)
> * Auto-post new (incoming) papers for each track into those dedicated channels so that awareness there is a new paper in a track isn't limited to only the track EiC.
> * Have multiple track EiCs.

> I'd like to advocate for the track co-leader model. @danielskatz is absolutely right about the need to take a break at times. Can we launch without this but be open to adding a co-lead after evaluating the workload?

I agree with all these suggestions, and I'm good with launching with a single person but adding another afterwards. It  definitely makes sense to have a second track EIC, but initially I think just being *slightly* less responsive (i.e., a day or two) is not that big of a deal.

arfon:
Yeah, it's a very reasonable concern! A couple of thoughts:

- We currently don't really receive submissions in these areas. If we did, then I would expect us to introduce new tracks to cover these areas. 
- I'd like us to avoid introducing an 'other/miscellaneous' track (which would potentially be one place topics/subject areas that don't fit elsewhere) if we can. But that is one possible alternative. 

What do you think @oliviaguest?

jedbrown (commented):
Some overlap is unavoidable, but I want to make sure it's recognized how much overlap (and lots of important sofware falls into this overlap) there is between, say @danielskatz's `Modeling and Simulation [Physical Sciences]` and the computational engineering in @kyleniemeyer as well as computational geosciences in @kthyng's sections. As long as everyone is clear that there will be no meaningful distinction between which track these papers fall into, I think that's okay, but it should feel like a low-stakes decision for authors.

Also, @gkthiruvathukal's `Data Science: Modeling and Simulation [Physical Sciences]` isn't really a distinct category. (I think it's harmful that some CS departments/researchers have conjured a false dichotomy that too often manifests as "doing X without domain knowledge in X", but that's a tangent for another time.)

oliviaguest:
Hmmm. Why not then not put them anywhere for now, and introduce them with their new track?

arfon:
> Why not then not put them anywhere for now, and introduce them with their new track? I feel like I shouldn't be singled out for areas I'm not remotely related to, @arfon, as that makes me/my track the miscellaneous category in everything but name.

A great point, and to be clear, I definitely don't want to ask you to lead a track that feels wrong/inappropriate/unsuitable.

This discussion has convinced me that we probably _should_ have a miscellaneous track, but perhaps not staff it with a track EiC yet. We could add all of the subject areas that we think don't really fit anywhere logically today, and if we do start to receive submissions in these domains, then, as you say, that would be good signal that we need to start adding editorial capacity in these areas.

oliviaguest:
Oh, cool. I will need to check with my current assigned areas which fit my expertise, so that I can make a coherent track. Shall I directly edit or PR the file itself?

arfon:
> Oh, cool. I will need to check with my current assigned areas to see which fit my expertise, so that I can make a coherent track. Shall I directly edit or PR the file itself?

Feel free to make suggestions in comments and/or push commits to the PR directly.

arfon:
> Oh, cool. I will need to check with my current assigned areas to see which fit my expertise, so that I can make a coherent track. Shall I directly edit or PR the file itself?

@oliviaguest – on the above ☝️, I just wanted to clarify that the editorial track will hopefully look pretty close to the subject areas we discussed earlier in [this Google doc](https://docs.google.com/document/d/12vowUKTfsPiPMg6Gny2g9YTLLISZiL0yRjCwjloeFDE/edit#). i.e., something along the lines of:

> **Social, Behavioral, and Cognitive Sciences (SBCS) (EiC: Olivia)**
> - Psychology
> - Political science
> - Urban planning
> - Economics
> - Neuroscience

Is that the sort of thing you're aiming for here? Also, if it would be helpful I'd be happy to book some time to chat more about how to allocate these subjects.

kthyng (commented):
Mine look fine though I am surprised "hydrology" isn't listed by name. I suppose it fits under something else with a broader name, but we do get quite a few hydrology submissions.

lib/tracks.yml
```diff
@@ -0,0 +1,400 @@
...
+      - Arts and Humanities (miscellaneous) [Social Sciences & Humanities]
+      - History [Social Sciences & Humanities]
+      - Language and Linguistics [Social Sciences & Humanities]
+      - Archeology (arts and humanities) [Social Sciences & Humanities]
+      - Classics [Social Sciences & Humanities]
+      - Conservation [Social Sciences & Humanities]
+      - History and Philosophy of Science [Social Sciences & Humanities]
+      - Literature and Literary Theory [Social Sciences & Humanities]
+      - Museology [Social Sciences & Humanities]
+      - Music [Social Sciences & Humanities]
```
arfon, line 260: Remove from track.

lib/tracks.yml
```diff
@@ -0,0 +1,400 @@
...
+      - Clinical Psychology [Social Sciences & Humanities]
+      - Developmental and Educational Psychology [Social Sciences & Humanities]
+      - Experimental and Cognitive Psychology [Social Sciences & Humanities]
+      - Neuropsychology and Physiological Psychology [Social Sciences & Humanities]
+      - Social Psychology [Social Sciences & Humanities]
+      - General Social Sciences [Social Sciences & Humanities]
+      - Social Sciences (miscellaneous) [Social Sciences & Humanities]
+      - Archeology [Social Sciences & Humanities]
+      - Development [Social Sciences & Humanities]
+      - Education [Social Sciences & Humanities]
```
arfon, line 296: Remove from track.

lib/tracks.yml
```diff
@@ -0,0 +1,400 @@
...
+      - Geography, Planning and Development [Social Sciences & Humanities]
+      - Health (social science) [Social Sciences & Humanities]
+      - Human Factors and Ergonomics [Social Sciences & Humanities]
+      - Law [Social Sciences & Humanities]
```
arfon, line 300: Remove from track.

lib/tracks.yml
```diff
@@ -0,0 +1,400 @@
...
+      - Library and Information Sciences [Social Sciences & Humanities]
```
arfon, line 301: Move to CSISM

lib/tracks.yml
```diff
@@ -0,0 +1,400 @@
...
+      - Linguistics and Language [Social Sciences & Humanities]
+      - Safety Research [Social Sciences & Humanities]
```
arfon, line 303: Remove from track.

arfon (commented):
Based on conversation with @oliviaguest

gkthiruvathukal (commented):
@arfon Should the test name "should have 7 top level tracks" be updated to "should have 8 top level tracks" to reflect the change on line 22?

PR #1110, opened 2022-07-22
Add conflict policy link to ethics guidelines

chartgerink:
This PR adds a link in the Ethics guidelines section of the about page, to further clarify the conflict policy of JOSS.

This is important as the PMC application will review whether these policies are there. This is my attempt to improve discoverability for the people who will be checking the application but also in general 😊


danielskatz:
great, thanks - I've sometime realized that the policy wasn't as easy to find as it should have been.

PR #1112, opened 2022-07-29
Capitalize "Need" in Statement of Need

dunnkers:
Change "Statement of need" into "Statement of Need" in the example paper on the documentation page. Without the capital "N", I got a warning from @editorialbot, saying:

> Failed to discover a `Statement of need` section in paper

This is actually the warning generated by:

https://github.com/openjournals/buffy/blob/912c2b0753c13ff750625c9d5ed49fd6fec53c2f/app/workers/repo_checks_worker.rb#L74

The code checks whether the text contains "Statement of Need".

Cheers ✓.

### DIFF
--- a/docs/submitting.md
+++ b/docs/submitting.md
@@ -141,7 +141,7 @@ Aside from toy problems and demonstrations, the majority of problems require
 efficient numerical tools, many of which require the same base code (e.g., for
 performing numerical orbit integration).
 
-# Statement of need
+# Statement of Need
 
 `Gala` is an Astropy-affiliated Python package for galactic dynamics. Python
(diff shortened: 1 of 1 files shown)


jedbrown:
I would expect the `/i` in the referenced regex would be a case-insensitive match. Also, I think our style (based on APA) prefers sentence case in headings and thus `Statement of need` is preferred.

arfon:
Yeah. I think we should relax the regex here to be less case sensitive.

dunnkers:
Ah right, the regex is already configured for insensitive casing. Didn't see that! In that case something else must've triggered the warning for me. I'll investigate a bit further on what the actual cause was.

Closing for now. Thanks for your response.

PR #1132, opened 2022-09-25
Track EiC notifications

arfon:
🚧 Don't merge this until we've shipped the JOSS tracks 🚧 

This PR implements notifications for track EiCs (via email) when submissions are received *and* when a paper is moved between tracks.


app/models/paper.rb
```diff
@@ -429,10 +429,13 @@ def set_track_id(new_track_id)
 
   def move_to_track(new_track)
     return if new_track.nil?
+    old_track = self.track
     current_label = self.track.present? ? self.track.label : ""
```
arfon, line 433: @xuanxu – does this mean we decide whether to move tracks based on the the labels rather than the track actual ActiveRecord ID? Is there a reason for this?

xuanxu: Yes, when I did this the idea was to create/update the tracks info running rake tasks that could potentially delete all tracks and regenerate them with the updated info, and that would change the id of the tracks. I thought the name or label would be a more stable info. So here if the labels are the same the system deduces that there is no need to change track. Does it make sense?

PR #1138, opened 2022-10-03
Add requirement of free and open accounts

jmschrei:
This PR adds the requirement that submissions must include repositories hosted at locations that have free account creation without manual verification, e.g. BitBucket or GitHub. These changes are meant to convey that it is not okay to host the code at a place where outside users cannot contribute, or user accounts must be manually approved before they can open issues or pull requests, even if the code is free to access. In my opinion, barriers to user participation should be discouraged, and this should be explicitly stated as one of our guidelines.

Looking forward to the discussion.

### DIFF
--- a/docs/review_criteria.md
+++ b/docs/review_criteria.md
@@ -119,3 +119,7 @@ We ask that reviewers grade submissions in one of three categories: 1) Accept 2)
 As outlined in our author guidelines, submissions that rely upon a proprietary/closed source language or development environment are acceptable provided that they meet the other submission requirements and that you, the reviewer, are able to install the software & verify the functionality of the submission as required by our reviewer guidelines.
 
 If an open source or free variant of the programming language exists, feel free to encourage the submitting author to consider making their software compatible with the open source/free variant.
+
+### Where can I host the repository?
+
+We ask that repositories are hosted at a location that allows outside users to freely open issues and make pull requests, such as BitBucket or GitHub. Submissions will not be accepted if the repository is at a location where users must pay to open issues or make pull requests, or where accounts must be manually approved, regardless of if this approval is done by the owners of the repository or some other entity. 
--- a/docs/review_criteria.md
+++ b/docs/review_criteria.md
@@ -122,4 +122,4 @@ If an open source or free variant of the programming language exists, feel free 
 
 ### Where can I host the repository?
 
-Repositories must be hosted at a location that allows outside users to freely open issues and propose changes, such as GitHub, GitLab, Bitbucket, etc. Submissions will not be accepted if the repository is hosted at a location where where accounts must be manually approved (or paid for), regardless of if this approval is done by the owners of the repository or some other entity. 
+Repositories must be hosted at a location that allows outside users to freely open issues and propose code changes, such as GitHub, GitLab, Bitbucket, etc. Submissions will not be accepted if the repository is hosted at a location where where accounts must be manually approved (or paid for), regardless of if this approval is done by the owners of the repository or some other entity. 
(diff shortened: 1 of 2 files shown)


oliviaguest:
Thanks for this. I would also like some feedback that is clear and actionable for the submission I am handling here, whether it means grandfathering in or not: https://github.com/openjournals/joss-reviews/issues/4600. Especially since editors @danielskatz @samhforbes @arfon and I have discussed this, but I am still not sure what is the plan. Apologies if this is more my own confusion, but I doubt it, since there have been almost as many contradictory views so far as there have been discussions. 😅

docs/review_criteria.md
```diff
@@ -119,3 +119,7 @@ We ask that reviewers grade submissions in one of three categories: 1) Accept 2)
 As outlined in our author guidelines, submissions that rely upon a proprietary/closed source language or development environment are acceptable provided that they meet the other submission requirements and that you, the reviewer, are able to install the software & verify the functionality of the submission as required by our reviewer guidelines.
 
 If an open source or free variant of the programming language exists, feel free to encourage the submitting author to consider making their software compatible with the open source/free variant.
+
+### Where can I host the repository?
+
+We ask that repositories are hosted at a location that allows outside users to freely open issues and make pull requests, such as BitBucket or GitHub. Submissions will not be accepted if the repository is at a location where users must pay to open issues or make pull requests, or where accounts must be manually approved, regardless of if this approval is done by the owners of the repository or some other entity.
```
danielskatz, line 125:
```suggestion
We ask that repositories are hosted at a location that allows outside users to freely open issues and make pull requests, such as GitHub, GitLab, BitBucket, etc. Submissions will not be accepted if the repository is at a location where users must pay to open issues or make pull requests, or where accounts must be manually approved, regardless of if this approval is done by the owners of the repository or some other entity. 
```

danielskatz: adding GitLab, fixing casing in Bitbucket, changing to order by most common

danielskatz:
@jmschrei - this looks good to me, with the one small edit I suggested. Do you also think we need to change anything in the review criteria?

danielskatz:
@oliviaguest - I think if we make this change, you can say to the authors that we making the implied JOSS policy explicit, and they need to follow it. (This is my opinion - I think @arfon, as the EiC, should make the final decision.)

docs/review_criteria.md
jmschrei: I don't have an example where users might have to pay, but it seemed worth including in case some service decides to start charging for accounts or a new service pops up.

jmschrei:
> @jmschrei - this looks good to me, with the one small edit I suggested. Do you also think we need to change anything in the review criteria?

Yes, we will probably need to add a bullet to check that this is being followed. I can add that after we've come to a decision. I am traveling this week, so may be delayed in adding that. 

> @oliviaguest - I think if we make this change, you can say to the authors that we making the implied JOSS policy explicit, and they need to follow it. (This is my opinion - I think @arfon, as the EiC, should make the final decision.)

I sort of go the other way on this. This wasn't explicitly our policy when they submitted, and since their submission, several months have gone past. I would be very frustrated if, several months into a review, the policy changed and suddenly I had to do a lot of new work. But I don't feel that strongly about it.

arfon:
This looks like a good change to me. I'm inclined to agree with @jmschrei here that existing submissions are 'grandfathered' in – i.e., do *not* have to meet this policy. @oliviaguest – I think this helps you with your submission right?

danielskatz:
I don't think we need to formal say that old submissions need to meet this, but we have held old submission up until they did, so I think we can do so again.  Anyhow, @arfon, you can decide on how to handle this one particular submission, knowing my opinion and those of others ...

PR #1143, opened 2022-10-15
Update to new checkout version

ianfhunter:
Removes Node js upgrade warning.

There may be documentation to be changed also for users who are setting up their own action for paper generation (how I found the issue)


xuanxu (approved):
:+1: Thanks @ianfhunter!

PR #1152, opened 2022-10-31
Check twice per week during pre-review

arfon:
Updating our editor docs to make it clear that checking on papers, before they have reviewers assigned, is especially important.


danielskatz (approved):
Lgtm

PR #1160, opened 2022-11-25
Submitting.md: add a TiKZ example

pedramardakani:
Until now, readers were not informed about their access to the powerful TiKZ library _inside_ their own paper. This might lead them to using rasterized graphics inside their publication which cannot compete with the infinite resolution of vector graphics.

Also, some might try hacking different methods to produce the illustrations and inject them inside the 'paper.md' file with the markdown syntax. This leads to inconsistent typesetting inside the final paper.

With this commit, I've put a TiKZ example taken from the pgfmanual documentation for a demonstration. This example enjoys the same typesetting as the rest of the paper. For example, the figure caption has the same font as the body.

### DIFF
--- a/docs/submitting.md
+++ b/docs/submitting.md
@@ -204,6 +204,30 @@ and referenced from text using \autoref{fig:example}.
 Figure sizes can be customized by adding an optional second parameter:
 ![Caption for example figure.](figure.png){ width=20% }
 
+You may use the `TiKZ` package as well, here is an example taken from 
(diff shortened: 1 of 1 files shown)


pedramardakani:
Hi all 👋 

The `openjournals/inara` docker image compiled the paper successfully. Hopefully this will encourage using the TiKZ library inside the paper easily without the need to find a _hack_.

Best,
Pedram

Issue #1179, opened 2023-01-27
Update year in docs


darigovresearch:
Year says 2018 in footer https://joss.readthedocs.io/en/latest/index.html

![image](https://user-images.githubusercontent.com/30328618/215214478-898492ba-4af9-49c9-a1df-2ca67a728424.png)

ethanwhite:
This issue was fixed by ed63cbca9744157080e48f013bb828fbf1586ca6

PR #1180, opened 2023-01-27
Update year to 2023 in docs

darigovresearch:
Related to #1179


xuanxu (approved):
We'll probably end up removing the year, but for now this is an improvement, thanks @darigovresearch!

PR #1193, opened 2023-03-14
Make a paragraph in the README flow better

rikhuijzer:
In the old version (*"Perhaps, and in a perfect world we'd rather papers about software weren't necessary but we recognize that for most researchers, papers and not software are the currency of academic research and that citations are required for a good career."*) there seems to be a typo because it should probably be *"in a perfect world, we'd rather **[have that]** papers ..."*. This PR suggests to write the paragraph a bit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ The Journal of Open Source Software (JOSS) is an academic journal with a formal 
 
 ### Don't we have enough journals already?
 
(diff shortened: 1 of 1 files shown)


navidcy (approved):

danielskatz:
👋 @arfon - I think this seems reasonable to merge, but will leave it to you to decided

arfon:
Looks good. Thanks @rikhuijzer.

PR #1231, opened 2023-06-05
Pagination: migrate from will_paginate to pagy

xuanxu:
[will_paginate](https://github.com/mislav/will_paginate) entered maintenance mode. 
[Pagy](https://github.com/ddnexus/pagy) seems like the best alternative right now. It's fast, efficient and compatible with searchkick/elasticsearch


xuanxu (commented):
Working in production. 
Memory consuption seems to be lower 👍

PR #1241, opened 2023-07-06
codespell: workflow, config and some typos

yarikoptic:
But not sure about "superceded" since seems to be requiring model change -- thus not worth it? should be skipped?


xuanxu:
We want to keep the actions we have to mantain for the repo at a minimum, so prefer not to run codespell. I've fixed the typos in a separate commit. Thanks @yarikoptic!

PR #1242, opened 2023-07-06
On submission form - inform user to start typing

yarikoptic:
In my case, having seeing no options given (just an editbox), I just quickly typed "neuroimaging" and no alert was given but I failed to submit since error said that I should choose a valid subject. So I removed "neuroimaging" but no choices has come about. I searched the internet on what subjects it could be -- no hint was given.  Only when I started to type slowly "neu" choices start to come up.

Hence, I think for the sake of people smoother experience, I think the hint there should be extended


xuanxu (approved):
:+1: Thanks @yarikoptic!

PR #1251, opened 2023-07-24
Remove guard

xuanxu:
Rails 7 includes hot reloading, good enough for development work. Let's remove some dependencies.


xuanxu (commented):
CI build green

PR #1256, opened 2023-08-11
Retraction papers

xuanxu:
This PR adds the ability to create retraction notices and retract papers using the API.

### DIFF
--- a/app/controllers/dispatch_controller.rb
+++ b/app/controllers/dispatch_controller.rb
@@ -151,7 +151,7 @@ class DispatchController < ApplicationController
       @paper = Paper.find_by_review_issue_id!(params[:id])
 
       retraction_paper = Paper.new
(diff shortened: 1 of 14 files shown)


xuanxu (commented):
Run migrations after merging

arfon:
@xuanxu – @chartgerink is looking to better understand how all of this works (for our PMC submission). Is this a good summary?

Essentially we now:

- Publish a new short retraction paper (authors are the 'JOSS editorial team' or whatever/whomever we want them to be').
- Link the retraction notice paper back to the original (retracted) paper.
- Mark the original paper as retracted in the JOSS UI (and link to the retraction notice).

I think that's about it right? Are there other things that happen in the Crossref metadata?

xuanxu:
Yeah, that's it. 
The retraction paper is deposited with Crossref as any other paper, no extra special metadata.

PR #1271, opened 2023-09-08
juliacon update

lucaferranti:
@xuanxu @arfon I tried to merge main into the `juliacon` and went through the conflicts. It wasnt too bad, but I assume I messed up something. Is there an easy(ish) way to deploy the website (at least as static) locally and check it looks about right?

A few notes I took while going through the conflicts

- the `repository.yml` file had a rename `joss-review-testing --> joss-reviews`, could you comment on which one we should use in this case?
- the files `settings-test.yml`, `settings-development.yml` and `settings-production.yml` seemed identical. I updated the metadata and marked as `FIXME` the new fields that wasnt sure how to fill. Most of them I can figure out asking juliacon people, however the following I didnt know how to setup, ideas what the correct link looks like?
  - reviewers_url: "FIXME"
  - reviewers_signup_url: "FIXME"
  - reviewers_lookup_url: "FIXME"
  
- `editor-email` and `noreply-email` was set to `admin@theoj.org` (also in the old version), not sure this is correct.

This is as draft for now, since we need to first assemble the gh action to compile latex, meanwhile we can start to take a look.

cc @matbesancon


xuanxu:
Hi Luca, here are some quick responses, I'll take a more detailed look at the PR code and go back to you.

> Is there an easy(ish) way to deploy the website (at least as static) locally and check it looks about right?

You need to have installed Ruby 3.2.2 and have running locally elasticsearch and postgres. Then: 
running `bundle` once should install all code dependencies
running `bin/rails db:create` once should create the database
running `bin/rails s` should start a server (stop it with ctl+c) and have the site available at localhost:3000

> the `repository.yml` file had a rename `joss-review-testing --> joss-reviews`, could you comment on which one we should use in this case?

The value of that header string is not really important, other than to be visually informative, so is not needed to update it but to be correct I suggest you change it from `openjournals/joss-reviews` to `JuliaCon/proceedings-review`

> Most of them I can figure out asking juliacon people, however the following I didnt know how to setup, ideas what the correct link looks like?
>       
>       * reviewers_url: "FIXME"

This points to a place where reviewers are managed, in JOSS we have a dedicated app, you can left it blank or use https://juliacon.github.io/proceedings-guide/reviewer/ 

>       * reviewers_signup_url: "FIXME"

This is the url of the place where potential reviewers can sign up, in your case I think is: https://forms.gle/Zr9jNaWg45qkAWKT6

>       * reviewers_lookup_url: "FIXME"

This is a link to the list of reviewers, if you don't have a public list I guess this is enough: https://juliacon.github.io/proceedings-guide/reviewer/

>     * `editor-email` and `noreply-email` was set to `admin@theoj.org` (also in the old version), not sure this is correct.

Yes, heroku email addon is configured to allow that email address to send the internal emails

PR #1274, opened 2023-09-14
Dashboard: list query scoped submissions

xuanxu:
This PR adds a new option to the editor dashboard to list only query scoped submissions (filterable by track)

Captura de pantalla 2023-09-14 a las 12 21 17

### DIFF
--- a/app/models/paper.rb
+++ b/app/models/paper.rb
@@ -131,6 +131,7 @@ class Paper < ApplicationRecord
   scope :search_import, -> { where(state: VISIBLE_STATES) }
   scope :not_archived, -> { where('archived = ?', false) }
   scope :by_track, -> (track_id) { where('track_id = ?', track_id) }
+  scope :query_scoped, -> { where("labels->>'query-scope' IS NOT NULL") }
 
   before_create :set_sha, :set_last_activity
   after_create :notify_editors, :notify_author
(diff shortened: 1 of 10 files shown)


app/models/paper.rb
```diff
@@ -131,6 +131,7 @@ class Paper < ApplicationRecord
   scope :search_import, -> { where(state: VISIBLE_STATES) }
   scope :not_archived, -> { where('archived = ?', false) }
   scope :by_track, -> (track_id) { where('track_id = ?', track_id) }
+  scope :query_scoped, -> { where("labels->>'query-scope' IS NOT NULL") }
```
arfon, line 134: I think we should probably exclude those papers that the editor has already voted on right?

xuanxu: Sometimes I want to go back to one I've voted on days ago to see how the votes go, if they desappear from the list those papers will be difficult to find. As an editor I think is more useful to have the list of all the in progress query scoped papers, usually they are going to be less than a page. But maybe is just me, I don't know, WDYT?

arfon: Oh, good point. Perhaps we can find some way to show if you've already voted (but still show it in the list)?

arfon: Like we do here: https://github.com/openjournals/joss/blob/bee3621e6c2c6258cfa982ed47bf4fae1e38f58f/app/views/notifications/editor_scope_email.text.erb#L7

xuanxu (commented):
The list now show if/what you've already voted:

votes

PR #1282, opened 2023-10-08
Rails 7.1

xuanxu:


xuanxu (commented):
Code should be updated to pass when `raise_on_missing_callback_actions` is true

xuanxu:
Ready!

PR #1292, opened 2023-11-03
Update physics and astronomy field keywords to reduce false positives

dfm:
I've been finding that we've been getting a fair number of incorrect track assignments for general physics papers in the AASS track. This is because the labels "General Physics and Astronomy" and "Physics and Astronomy (misc)" were both being sent to AASS. This PR splits those field labels to assign "Physics" to the PE track and "Astronomy" to the AASS track.

cc @kyleniemeyer


arfon:
Looks good @dfm! @xuanxu – what is the process for re-importing the subjects? Can we just re-run the rake task?

xuanxu:
@arfon to re-import just run the `tracks:import:subjects` task. 
Before that you can run `tracks:import:dry_run` to make sure it'd work and get some info on what will be imported

arfon:
OK, those should be imported now @dfm

dfm:
Thanks @arfon and @xuanxu!!

PR #1314, opened 2024-02-13
website updates

lucaferranti:
@arfon @xuanxu 

btw it seems I cannot preview anymore the build, but heroku says build was successful, any idea why that may be?


arfon:
I think because the database is broken? (Waiting on some support from @xuanxu on this)

lucaferranti:
How preview works and a few more links to fix, so moving back to draft until I finish inspecting

lucaferranti:
supersed by #1329

PR #1318, opened 2024-03-01
Clarify author metadata check

RMeli:
I asked the following on Slack:

> Do the metadata on Zenodo need to match the paper exactly? Or the author list of the paper can be a subset of the authors on Zenodo?

and this PR changes the pre-publication steps according to the answers received.

My current understanding is that it is OK if the authors in the archive metadata don't match the authors on the paper as long as all the authors on the paper are included in the metadata (other contributors should be marked as such, if possible, instead of being marked as authors too).

### DIFF
--- a/docs/editing.md
+++ b/docs/editing.md
@@ -104,7 +104,7 @@ Pre-publication steps:
 - Proof-read the paper and ask authors to fix any remaining typos, badly formed citations, awkward wording, etc..
 - Ask the author to make a tagged release and archive, and report the version number and archive DOI in the review thread.
 - Check the archive deposit has the correct metadata (title and author list)
-  - It is _preferred_ that the metadata author list matches the paper author list
+  - In most situations, the two author lists should match. Authors and editors should review the two, and any differences need to be explained.
     - Other contbutors can be present (and they should be marked as such, if possible)
   - Check that the all authors of the paper are in the archive metadata
   - Eventually, ask for the reason why the two lists differ
--- a/docs/editing.md
+++ b/docs/editing.md
@@ -97,12 +97,17 @@ When a submission is ready to be accepted, we ask that the authors issue a new t
 Optionally you can ask EditorialBot to generate a checklist with all the post-review steps running the command: `@editorialbot create post-review checklist`
 
 Pre-publication steps:
+- (Optional) Run `@editorialbot create post-review checklist`
 - Get a new proof with the `@editorialbot generate pdf` command.
 - Download the proof, check all references have DOIs, follow the links and check the references.
   - EditorialBot can help check references with the command `@editorialbot check references`
 - Proof-read the paper and ask authors to fix any remaining typos, badly formed citations, awkward wording, etc..
 - Ask the author to make a tagged release and archive, and report the version number and archive DOI in the review thread.
-- Check the archive deposit has the correct metadata (title and author list), and request the author edit it if it doesn’t match the paper.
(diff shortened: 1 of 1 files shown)


docs/editing.md
```diff
@@ -97,12 +97,17 @@ When a submission is ready to be accepted, we ask that the authors issue a new t
 Optionally you can ask EditorialBot to generate a checklist with all the post-review steps running the command: `@editorialbot create post-review checklist`
 
 Pre-publication steps:
+- (Optional) Run `@editorialbot create post-review checklist`
```
RMeli, line 100: This is a bit repetitive since it's written just above, but if someone goes too quickly to the list of steps, this suggestion can be easily missed.

docs/editing.md
```diff
@@ -97,12 +97,17 @@ When a submission is ready to be accepted, we ask that the authors issue a new t
...
 - Get a new proof with the `@editorialbot generate pdf` command.
 - Download the proof, check all references have DOIs, follow the links and check the references.
   - EditorialBot can help check references with the command `@editorialbot check references`
 - Proof-read the paper and ask authors to fix any remaining typos, badly formed citations, awkward wording, etc..
 - Ask the author to make a tagged release and archive, and report the version number and archive DOI in the review thread.
-- Check the archive deposit has the correct metadata (title and author list), and request the author edit it if it doesn’t match the paper.
+- Check the archive deposit has the correct metadata (title and author list)
+  - It is _preferred_ that the metadata author list matches the paper author list
```
emdupre, line 107:
```suggestion
  - It is _strongly preferred_ that the metadata author list matches the paper author list exactly. Confirm that any discrepancies between the two listings are intentional and that the authors can clearly justify this decision. 
```

emdupre: Thanks for making this suggestion ! I think we could consolidate into one bullet point using something like this revised language. Let me know what you think !

RMeli: I think it's a good idea. But is it really a strong preference for the author list to match exactly? For me there is a difference between core authors/developers and other contributors for a paper, but I don't see why in Zenodo/other metadata it would be preferred to potentially remove contributors to match the paper author list. I might be missing something, but is it JOSS expectation that all contributors, no matter how small their contribution, should be authors of the paper?

kyleniemeyer: @RMeli yes, for most projects the author list and Zenodo author/contributor list should match. Any other situation needs to have a valid rationale, and probably distinguish between contributors in Zenodo. (In most cases, people remove the extraneous authors from Zenodo—which frequently get brought in accidentally—rather than add people to the paper.)

So, perhaps rather than making it a preference, strong or otherwise, we can just say that: In most situations, the two author lists should match. Authors and editors should review the two, and any differences need to be explained.

RMeli: > In most cases, people remove the extraneous authors from Zenodo—which frequently get brought in accidentally—rather than add people to the paper.

That's a bit what I'm concerned about, but probably it's just my point of view. For me Zenodo metadata should contain all contributors (and I agree that ideally the distinction between contributor and author should be used), so in my mind the requirement should be that the Zenodo author list has to contain all the authors of the paper, but can have more people (that contributed to the project in minor ways).

> So, perhaps rather than making it a preference, strong or otherwise, we can just say that: In most situations, the two author lists should match. Authors and editors should review the two, and any differences need to be explained.

That's a good idea, I think. It makes clear that clarifications are in order, but does not recommend a specific action that remove contributors (likely added automatically because they are in the commit history).

PR #1341, opened 2024-05-22
fix examples for ref and autoref

### LINKED ISSUE #1340: issue about cross references
I use the following cross-references according to [the docs of joss](https://joss.readthedocs.io/en/latest/paper.html#tables-and-figures). But why dose it result in ???

```md
A concise comparison ... frameworks provided in \autoref{frameworks}, such as ....

: Comparison ... frameworks. []{label="frameworks"}
```

The examples provided by the docs are not displayed in a right format also.

sneakers-the-rat:
Fix: https://github.com/openjournals/joss/issues/1340

looks like this wasn't working even before the transition to myst: https://web.archive.org/web/20230925202053/https://joss.readthedocs.io/en/latest/submitting.html#internal-references

left in the mention of the markdown-style labels even if they aren't working for now bc i assume that will be fixed.

Before: 
Screenshot 2024-05-21 at 8 19 26 PM

After:
Screenshot 2024-05-21 at 8 18 52 PM

### DIFF
--- a/docs/paper.md
+++ b/docs/paper.md
@@ -113,15 +113,39 @@ The goal of Open Journals is to provide authors with a seamless and pleasant wri
 
 Tables and figures can be referenced if they are given a *label* in the caption. In pure Markdown, this can be done by adding an empty span `[]{label="floatlabel"}` to the caption. LaTeX syntax is supported as well: `\label{floatlabel}`.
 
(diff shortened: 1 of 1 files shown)


xuanxu (approved):
LGTM

Consulted with @tarleb; the problem is that the labels mentioned in the docs do work, but only when going via ConTeXt instead of LaTeX. We expected a switch to ConTeXt soon, so this wasn't implemented for LaTeX. Maybe is better to fix the LaTeX output rather than changing the docs.

xuanxu:
@arfon about ConTeXt vs LaTeX: any word on what you think we'll use in the future?

sneakers-the-rat:
ping on this, still broken

danielskatz:
👋 @xuanxu - How should we deal with this PR?

Issue #1346, opened 2024-06-14
`papers/by/AUTHOR` page doesn't work for names containing period characters


kandersolar:
On the page for published papers, the list of author names in the side bar is hyperlinked to a page listing all publications by each author.  However, that page seems broken (HTTP 406) for author names containing a period, at least on Chrome -- I observe no issues with Firefox.

Here is an example:

- Paper page: https://joss.theoj.org/papers/10.21105/joss.06714
- Broken link: https://joss.theoj.org/papers/by/James%20P.%20Ewen

Deleting the "." in the URL gets it working for me.

sneakers-the-rat:
Can replicate in local development server. getting a 306 Not acceptable, "ActionController::UnknownFormat in PapersController#filter" right here: https://github.com/openjournals/joss/blob/4cd24daa5176eb8b4f6d7ecb0a93133b9cddc0ca/app/controllers/papers_controller.rb#L136

Chrome doesn't URL encode the `.` as `%2E` while firefox does.

PR #1356, opened 2024-08-15
Improve development environment & instructions.

sneakers-the-rat:
Was trying to help out and debug https://github.com/openjournals/joss/issues/1346
since it seems like a pretty big problem, but had a tough time figuring out how to actually get a development environment set up.

I also was having a tough time figuring out the models and so i added the `annotations` gem to create model schemas as comments above the models, i've found this pretty dang helpful while working on mastodon.

Also added some seed data when seed is run in a development environment bc i also find this super helpful :). I don't know if there's a good way to use the factories from spec to populate this, but that obvi would be better - took me awhile to figure out how to create a paper until i looked there.

not sure why the schema removed the `serial` id, happened when i ran the migration.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -52,6 +52,7 @@ group :development, :test do
   gem 'rspec-rails', '~> 6.1.0'
   gem 'selenium-webdriver'
   gem 'webmock'
+  gem 'annotate'
 end
 
 group :test do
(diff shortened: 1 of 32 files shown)


xuanxu:
In general we're against adding dependencies unless strictly needed. All the information added by the `annotate` gem can be found in the [schema file](https://github.com/openjournals/joss/blob/main/db/schema.rb).

The [updated docs in the readme](https://github.com/openjournals/joss/pull/1356/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5) are fine though, can you open a PR with only that?

sneakers-the-rat:
Dang even for dev dependencies? Thats a shame.

PR #1357, opened 2024-08-15
fix authors with periods in names

sneakers-the-rat:
Fix: https://github.com/openjournals/joss/issues/1346

The problem was that names with periods in them were interpreted as being a file extension (perhaps reasonably?) by the router. So clicking on `"Author T. Lastname"` in a paper would be interpreted by rails as `{author: "Author T", format: " Lastname"}`

So i added a constraint that matches any string excluding the three formats that it looks like there were responders for and added tests - don't know if i put the tests in the right spot, there weren't any tests for the `papers/by/:author` route so i just put it with the controller.

edit: i find ruby's formatting norms to be bewildering and bad, and i didn't notice a linter config in this repo, so sorry if my indentation is weird, feel free to request edits.


arfon:
Oooo, nice! Thanks for spotting (and fixing) this.

PR #1358, opened 2024-08-20
Development Environment Docs & Improvements

sneakers-the-rat:
Redo: https://github.com/openjournals/joss/pull/1356 without model annotations

This PR:
- Add more docs on setting up development environment
- Don't validate review repository URLs on creating `Paper` objects if in the development environment so that it's possible to create fake papers/reviews for testing
- Add a "paper" type to dev environment config
- Create some basic seed data for dev environment. 

Questions:
- is the readme the right place? i also notice https://github.com/openjournals/joss/blob/main/docs/installing.md which seems more like if you were to want to do a secondary deployment for a different journal, but just checking
- I only added enough seed data to be able to test https://github.com/openjournals/joss/pull/1357 , so eg. I didn't put a full paper body in there. is there some way to use the factory classes in the `spec` directory like https://github.com/openjournals/joss/blob/main/spec/factories/papers_factory.rb in `db/seeds.rb`? I don't know enough about ruby environments, wasn't sure how to call that, but that would probably be preferable to use those, so then we could create a full seed dataset with all the types of models from the same sources that the tests use. If that is possible/desirable, i can follow on with another PR for that, seems like it would be a nice thing to do both for people checking this out if they wanted to do a secondary deployment and also for quick debugging purposes.

### DIFF
--- a/app/models/paper.rb
+++ b/app/models/paper.rb
@@ -146,7 +146,7 @@ class Paper < ApplicationRecord
   validates :kind, inclusion: { in: Rails.application.settings["paper_types"] }, allow_nil: true
   validates :submission_kind, inclusion: { in: SUBMISSION_KINDS, message: "You must select a submission type" }, allow_nil: false
   validates_format_of :repository_url, with: /\Ahttps?:\/\//i, on: :create, message: "Repository URL is missing the protocol segment (http/https)"
-  validate :check_repository_address, on: :create, unless: Proc.new {|paper| paper.is_a_retraction_notice?}
+  validate :check_repository_address, on: :create, unless: Proc.new {|paper| Rails.env.development? || paper.is_a_retraction_notice?}
 
   def notify_editors
     Notifications.submission_email(self).deliver_now unless self.is_a_retraction_notice?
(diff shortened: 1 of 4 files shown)


app/models/paper.rb
```diff
@@ -146,7 +146,7 @@ class Paper < ApplicationRecord
   validates :kind, inclusion: { in: Rails.application.settings["paper_types"] }, allow_nil: true
   validates :submission_kind, inclusion: { in: SUBMISSION_KINDS, message: "You must select a submission type" }, allow_nil: false
   validates_format_of :repository_url, with: /\Ahttps?:\/\//i, on: :create, message: "Repository URL is missing the protocol segment (http/https)"
-  validate :check_repository_address, on: :create, unless: Proc.new {|paper| paper.is_a_retraction_notice?}
+  validate :check_repository_address, on: :create, unless: Proc.new {|paper| Rails.env.development? || paper.is_a_retraction_notice?}
```
arfon, line 149: 😍

arfon (approved):

arfon:
Thanks @sneakers-the-rat !

PR #1362, opened 2024-09-02
Add ROR affiliation links to example paper

### LINKED ISSUE #1353: JOSS and ROR
If JOSS asked authors to use [RORs](https://ror.org) to indicate their affiliations, we could pass them to Crossref and be a better member of the community, and it would also be easier for us to understand our authors.

We could also use RORs for editors and reviewers.

cthoyt:
Closes #1353

In https://github.com/openjournals/inara/pull/72, I added ROR support to Inara such that ROR identifiers can be added to affiliations. They therefore show up with nice icons in the PDF exports, inside the Crossref XML output, and in the JATS output.

As suggested by Arfon in https://github.com/openjournals/inara/pull/72#issuecomment-2325088069, I've updated the JOSS documentation to give users a hint on how they can take advantage of this themselves. This PR specifically does the following:

- [x] add ROR to example paper
- [x] update paper guidelines

As mentioned in https://github.com/openjournals/inara/pull/72#issuecomment-2329888306, there are a few deployment things that need to be done before this is live, but I think we're ready to merge this anyway :)

### DIFF
--- a/docs/example_paper.md
+++ b/docs/example_paper.md
@@ -31,6 +31,7 @@ authors:
 affiliations:
  - name: Lyman Spitzer, Jr. Fellow, Princeton University, USA
    index: 1
+   ror: 00hx57361
  - name: Institution Name, Country
    index: 2
  - name: Independent Researcher, Country
--- a/docs/paper.md
+++ b/docs/paper.md
@@ -98,6 +98,29 @@ authors:
   <!--     given-names: 瀧 -->
   <!--     surname: 立花 -->
 
+## Affiliations
+
+Each affiliation requires an `index` and `name`.
+
+Optionally, the Research Organization Registry (ROR) identifier for the top-level
+organization can be annotated with the `ror` key. Note that ROR does not include
+departments in its [scope](https://ror.org/registry/#scope-and-criteria-for-inclusion),
+so ROR annotations are typically made to the top-level organization.
+
+```yaml
+authors:
+  - name: Albert Krewinkel
+    affiliation: [ 1, 2, 3 ]
+
+affiliations:
+  - index: 1
+    name: Open Journals
+  - index: 2
+    name: Pandoc Development Team
+  - index: 3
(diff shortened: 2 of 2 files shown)


danielskatz:
Thanks @cthoyt - I think this looks good.

But I don't think merging this makes sense at this point until we also do something with this information that we'll now sometimes get in the .md papers.

Specifically:
1. We want this to appear in the built PDF
2. We want this to appear in the built JATS file
3. (I'm less sure about this) - We want this to be in the metadata (xml file) we submit to Crossref

danielskatz:
Another question is if we want this to be optional or required. I lean towards required, but can see arguments both ways.

cthoyt:
@danielskatz the PR I mentioned in the top accomplishes all of the things you requested - it adds ROR links in PDF, it updates Crossref XML generation following their form for annotating affiliation identifiers, and this ultimately gets pulled into JATS. Please take a look at https://github.com/openjournals/inara/pull/72, and if you have any other ideas on functionality, please let me know!

Unfortunately, not all affiliations can be annotated with ROR, so I wouldn't enforce a requirement. Though, it's great to have when possible.

danielskatz:
Ok, sorry - I didn't see that this was connected to another PR - I was expecting all of this to happen in a single PR.  My mistake...  I'll take a look at the whole thing in a little bit.

danielskatz:
Ah, I now see that @arfon and @tarleb have reviewed the other PR - yes, so once that is merged, this seems good to me to be merged to.

Thanks very much for doing this!!

danielskatz (approved):

xuanxu (approved):
:+1: LGTM

cthoyt:
Since https://github.com/openjournals/inara/pull/72 has been merged, this is ready to be merged! @arfon would you please do the honors?

cthoyt:
thanks @xuanxu !

PR #1364, opened 2024-09-04
Add codespell support (config) and make it fix few typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

This is a reincarnated prior attempt 
- https://github.com/openjournals/joss/pull/1241

from a year ago, which shows that
- new typos got detected (and/or introduced)
- it is inconvenient to look for new typos without codespell config present so I had to copy it from that older PR
- there is a typo in the model (`superceded`) which if already permeated to other places -- I guess ok to keep around , but otherwise ideally should have been fixed.

@xuanxu mentions

> We want to keep the actions we have to mantain for the repo at a minimum, so prefer not to run codespell.

which is I guess ok, but is there another level of assistance/automation to look adding it to?

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,6 @@
+[codespell]
+# Ref: https://github.com/codespell-project/codespell#using-a-config-file
+skip = .git*,*.pdf,*.svg,vendor,*.lock,*.css,.codespellrc
+check-hidden = true
+ignore-regex= .*view=FitH.*
+ignore-words-list = te,aas,nwo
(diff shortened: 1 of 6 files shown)


sneakers-the-rat:
This seems like a minimal maintenance burden thing, and since it is our publicly facing site (rather than correcting typos in code/comments that only programmers would see) it makes sense to me to have. 

Maybe one thing that might ease anxiety around maintenance burden might be to scope it to just the parts of the package that contain public-facing strings like the erb templates? But then again there are probably some of those throughout the ruby and js too so maybe thats more complicated than it's worth.

PR #1368, opened 2024-10-06
Remove the 'waitlisted' label if it exists

arfon:
When moving between pre-review and review, we should remove the 'waitlisted' label if present.


arfon:
We don't have tests to easily cover this so going to merge – feels relatively low risk.

Issue #1377, opened 2024-10-23
improvements to TOC


danielskatz:
Some suggestions/requests for the [JOSS toc](https://joss.theoj.org/toc)  :

1. I find the dates in the issue pages (e.g., https://joss.theoj.org/toc/issue/102) quite confusing - the dd-mm-yyyy format would be clearer if it was yyyy-mm-dd.

2. The page's html title is just "Journal of Open Source Software" without saying "table of contents"

3. I don't see any link to this page from https://joss.theoj.org or https://joss.theoj.org/papers/published or https://joss.theoj.org/about

sneakers-the-rat:
oh wow i also didn't know this existed. 
(1) big yes on this, very agreed.
(2) Also seems like a good idea - also elsewhere we put "Journal of Open Source Software" first before any subtitle (eg. `Journal of Open Source Software: My Paper Title` on a paper page) which makes the title sort of useless, what about in addition to adding "Table of Contents" to the ToC page, we move the journal title to the end? like `My Paper Title - Journal of Open Source Software`. I think that's a pretty universal pattern

(3) it looks like there is one link but it's way down in the corner on the about page.

<img width="1680" alt="Screenshot 2024-10-23 at 12 19 36 PM" src="https://github.com/user-attachments/assets/0c2d419a-fbfb-4029-b25e-08a4da6f5b25">

i like the papers view, and i sort of like the free-flowing presentation without volumes/issues, but the toc view is super nice to see an abbreviated overview of papers through time. From a navigational point of view, it makes sense to me to want to go to "papers" to see the papers, and so it fees like "toc" would be just a different way of presenting the papers - so maybe we could put a link button up in this open space?

<img width="1078" alt="Screenshot 2024-10-23 at 12 27 04 PM" src="https://github.com/user-attachments/assets/aa2752d7-5383-4ffb-8e36-a4ecdd8d6aea">

while we're talking about style improvements, can i throw on the pile two more
(4): the alignment here is giving me a bit of an optical illusion effect, could this be a grid?

<img width="990" alt="Screenshot 2024-10-23 at 12 29 59 PM" src="https://github.com/user-attachments/assets/768b5ad0-28c7-4459-94c7-4c05f37c7b05">


[...]

PR #1378, opened 2024-10-23
Make ToC dates y-m-d

sneakers-the-rat:
Partial fix of: https://github.com/openjournals/joss/issues/1377

Just makes the dates `'%Y-%m-%d'` in the toc listings :). this is the only place that uses `'%d-%m-%Y'` on the site according to regex :)


danielskatz (approved):

xuanxu (approved):
:+1:

PR #1379, opened 2024-10-23
Titles in ToC

sneakers-the-rat:
Partial fix of: https://github.com/openjournals/joss/issues/1377

Cool! learned how to use `yield` and `content_for` in rails templates.

Screenshot 2024-10-23 at 1 19 19 PM

Should keep page title the same if there is no title provided, and otherwise be `{title} - Journal of Open Source Software`

Put the page name first because i don't think i'm alone in having my tabs often looking like this

Screenshot 2024-10-23 at 1 20 21 PM

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -11,7 +11,7 @@
     gtag('config', '<%= setting(:google_analytics) %>');
   </script>
   
-  <title><%= setting(:name) %></title>
+  <title><% if content_for(:title) %><%= yield :title %> - <%= setting(:name) %><% else %><%= setting(:name) %><% end %></title>
   <%= stylesheet_link_tag    'application', media: 'all' %>
   <%= javascript_importmap_tags %>
   <meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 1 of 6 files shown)


danielskatz:
I wonder if these should be "JOSS: {title}"? I kind of like the JOSS ones to be together in my menu/list...

sneakers-the-rat:
we can do whatever we want!

I'm looking at just my tabs right now, and all of them are `{page title} - {site title}`

Screenshot 2024-10-23 at 1 34 34 PM

So that is the norm, and then the favicon handles the grouping -- maybe your tab hygeine is better than mine ;)

Another thing is SEO, where having the full name in the page title is good (SEO can be a bit of mythology sometimes, so it may not matter that much if it's also in the header metadata), and `JOSS: {page title} - Journal of Open Source Software` is a little awkward.

but ya not strongly opinionated, just a frequent too-many-tabs-open-haver

danielskatz:
I hate to admit this, but I don't use tabs, I use windows, and I choose them from a menu that lists them in alphabetical order...

xuanxu (approved):
Nice improvement!

Issue #1385, opened 2024-11-01
GitHub badges for review (and papers edited) counts


arfon:
Currently it's possible to create a GitHub badge for a README with the DOI for a paper. What if we made it possible for a reviewer to link easily to their reviews on JOSS? We already have a URL that links to the reviews for a reviewer (e.g., https://joss.theoj.org/papers/reviewed_by/@speth) and the same for editors (https://joss.theoj.org/papers/edited_by/@arfon)

sneakers-the-rat:
I have a free hour to procrastinate before getting back into writing mode and this seems doable. I'm surprised we don't have a `Reviewer` model in the repo. that would be sort of the natural way to do it so it would mirror the `Paper#status` method, but since i get the sense that we don't want to touch this repository that much, I can just make a `badges` controller and do it that way. 

The `reviewed_by` route is currently going to the `papers#filter` controller, and it seems wrong to tack the badge on beneath that in the controller.

Issue #1388, opened 2024-11-01
Make `Reviewer` model?


sneakers-the-rat:
Opening this as a question - working on [making badges](https://github.com/openjournals/joss/issues/1385) right now, and reading the minutes from the meeting today (sorry i just never know when these are!), I am wondering if we want to make a `Reviewer` model? 

It makes sense to exist semantically, since `Reviewer`s exist! We could start with just having the single field of the gh username and appropriate relations to the other person models because a person can be a reviewer and an editor and an author, and then iterate on that as needed :)

Would be happy to draft this.

PR #1389, opened 2024-11-02
Reviewer Badges :)

sneakers-the-rat:
Fix: https://github.com/openjournals/joss/issues/1385

here she is:

https://github.com/user-attachments/assets/3bcab79c-cecc-4a90-afef-89076513e5e1

and on mobile:

https://github.com/user-attachments/assets/fb0d9e92-050e-4574-85f2-d9141f95efe7

OK this was more complicated than i thought because it turns out you need to like calculate the widths of all the characters for a given font to make an SVG badge, but i did it because i have wanted to have something like this in my own websites without having to use shields.io or w/e. Elsewhere the SVGs are just hardcoded, but I thought it would be nice to generate them for any key and color because we might want to use them for other things :)

I basically transcribed and simplified shields.io's code, it's mostly self explanatory except don't ask me what's going on with those scaling parameters. I grabbed an array of codepoint widths and made it a constant, use that to calculate the width, and the rest is in the controller.

I tried to reuse as much as i could from the rest of the site, but it looks like we're not using sass variables so i just hunted out values that seemed like they were the right ones to use. 

The badges will be at `joss.theoj.org/badges/reviewed_by/:username`. I imagine re-rendering them from scratch every time is pointlessly expensive, so plz lmk how caching works here because they should only rarely change, but it's hard to detect the review count being stale because we don't have a `Reviewer` model.

I am holding off on tests for now until i get signoff on the basic approach, and then i'll write em!


arfon:
Hey! Thanks for getting going on this @sneakers-the-rat ⚡. This is definitely the sort of thing I'm thinking about. One thought, do you think as part of this we might want to consider consolidating the other SVG rendering (e.g., https://github.com/openjournals/joss/blob/c447b2a349c3615b0225c57dee741c1494da943c/app/controllers/papers_controller.rb#L307-L321) into a common controller?

sneakers-the-rat:
I think it would make sense to do that, but didnt want to change it all until I got a nod to go ahead :)

sneakers-the-rat:
reviewing my open PRs and thought i'd ping on this? 

The only outstanding question i think is caching, and from what I can tell from the docs, rails should automatically cache rendering of the template for a given set of parameters. since the parameters given to the template are all generic across reviewer identity (i.e. they just have the text of the badge and the numbers) that should kick in. Ideally we would also cache the badge per reviewer so we don't need to do a db query at all, but doing that would require us to [add reviewers to the schema](https://github.com/openjournals/joss/issues/1388) so we can do cache invalidation when a reviewer does another review, and that should probably be a different PR. I also think swapping the existing badges to this controller should probably be another PR.

Anything blocking this?

arfon:
:zap: thanks again @sneakers-the-rat – just took this for a spin in production and it's looking great!

sneakers-the-rat:
Ok wait but mine is wrong I think? May be different filtering rules on the paper display vs the count
https://joss.theoj.org/papers/reviewed_by/@sneakers-the-rat

arfon:
Fixed in https://github.com/openjournals/joss/pull/1397

PR #1390, opened 2024-11-02
fix tests: temporarily add net-pop to deps

sneakers-the-rat:
The tests are broken :(

We are getting this bug: https://github.com/ruby/net-pop/issues/26

so until we can update to 3.3.4, this is a temporary workaround that forces the dependency resolution to run by installing from github rather than ruby gems


sneakers-the-rat:
wait no it was just my tests on my fork.. which should be up to date? but over here it looks like they're running so sorry for the spam

PR #1392, opened 2024-11-02
[style] fix orcid button alignment, less awkward intermediate-width menu

sneakers-the-rat:
currently the orcid button is out of alignment:

Screenshot 2024-11-01 at 11 35 52 PM

and on the mobile menu is has no horizontal padding. the underlines also extend too low into the neighboring item

Screenshot 2024-11-01 at 11 36 08 PM

---

So this pr aligns it

Screenshot 2024-11-01 at 11 38 21 PM

and since the horizontal line animation only really works on a horizontal menu, and the only time it's seen is on intermediate widths on desktops/laptops (since tablets/phones don't hover) where we would be scooting a mouse up and down, so i replaced it with a small size animation that has vertical rather than horizontal flow

and also added padding to the orcid button and prevented it from inverting color on hover since it was the only thing to do that

https://github.com/user-attachments/assets/55f37f22-bc95-42cc-bdd9-b952141a68ae

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -808,8 +808,10 @@ $btn-primary-border: darken($btn-primary-bg, 5%) !default;
 }
 
 .navbar-nav {
+    align-items: center;
+
     @include media-breakpoint-down(md) {
-        display: inline-block;
+        align-items: flex-end;
     }
 }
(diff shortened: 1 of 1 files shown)


xuanxu (approved):
:+1:

PR #1405, opened 2025-01-15
comment out compiling tools checkpoint

lucaferranti:
The box pointed to JOSS documentation, which mentions markdown, this confused a few people, since we use latex (and explicitly say you shouldn't have markdown).

Ideally, we would point to our compiling instructions, but those need some ironing, so I'm commenting out the box for now


xuanxu (approved):
:+1:

PR #1411, opened 2025-02-28
Add note about using AI in "Review criteria"

jromanowska:
I have been asked by reviewers whether we have a policy on AI-generated code. Perhaps we should just add a short clarifying text in our docs. Here's a suggestion

### DIFF
--- a/docs/review_criteria.md
+++ b/docs/review_criteria.md
@@ -125,4 +125,4 @@ Repositories must be hosted at a location that allows outside users to freely op
 
 ### Using AI to generate code
 
-JOSS does not require the authors to declare whether they have used AI to generate code. The AI tools are now often integrated in programming environments and it would be difficult to check for this. However, the authors should always check the code and be able to change it and adapt to reviewer's comments.
+JOSS does not have a policy regarding the use of AI/LLMs yet, but authors are responsible for understanding and explaining submitted code and its provenance, and should respond in good faith to reviewer questions about LLM use as they would with any other topic.
--- a/docs/review_criteria.md
+++ b/docs/review_criteria.md
@@ -122,3 +122,7 @@ If an open source or free variant of the programming language exists, feel free 
 ### Where can the repository be hosted?
 
 Repositories must be hosted at a location that allows outside users to freely open issues and propose code changes, such as GitHub, GitLab, Bitbucket, etc. Submissions will not be accepted if the repository is hosted at a location where accounts must be manually approved (or paid for), regardless of if this approval is done by the owners of the repository or some other entity. 
+
+### Using AI to generate code
+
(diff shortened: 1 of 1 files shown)


logological:
The proposed text could do with some minor stylistic/grammatical improvements:

> JOSS does not require authors to declare whether they have used AI to generate code. (AI tools are now often integrated in programming environments and it would be difficult for editors to check whether they have been used.) However, authors should always check their code and be able to change it and adapt it to reviewers' comments.

jedbrown:
I agree with @sneakers-the-rat, and would go further, that people should know (and thus be defensive toward) the use of Copilot and similar products can automate what would assuredly be copyright violation if a human did it. For example, it reproduces substantial portions of SuiteSparse (an LGPL library) with copyright stripped and only trivial cosmetic changes ([article from 2022](https://spectrum.ieee.org/ai-code-generation-ownership)), and this question of whether slight obfuscation is defense against copyright infringement is the subject of [ongoing class-action litigation](https://www.saverilawfirm.com/our-cases/github-copilot-intellectual-property-litigation). (Obfuscation is not a defense against copyright infringement by a human, which is why [clean-room design](https://en.wikipedia.org/wiki/Clean-room_design) exists.) Companies are banking on use of such products being sufficiently ubiquitous they can obtain a legislative solution if courts rule against them. They may be successful, and yet I object on ethical grounds.

Using products like Copilot to steal community work with plausible deniability is going to rack up technical debt while starving the high-quality libraries of resources to do original development and necessary maintenance. I don't think JOSS should encourage practices that harm the community JOSS was created to serve. There is a reality that many people use these products, often in lighter ways, and a blanket ban is not enforceable. But we can articulate that the products are ethically fraught, legally questionable, and threaten the goodwill of the community.

sneakers-the-rat:
for the sake of this PR, since this is a contentious topic and reviewers/authors are asking about it, maybe something like "JOSS does not have a policy re: use of LLMs yet, but authors are responsible for understanding and explaining submitted code and its provenance, and should respond in good faith to reviewer questions about LLM use as they would with any other topic."

danielskatz:
> for the sake of this PR, since this is a contentious topic and reviewers/authors are asking about it, maybe something like "JOSS does not have a policy re: use of LLMs yet, but authors are responsible for understanding and explaining submitted code and its provenance, and should respond in good faith to reviewer questions about LLM use as they would with any other topic."

I like this. I feel like responsibility is a key concept here.

jromanowska:
Thank you for the discussion! Yes, I also feel like the responsibility of authors should be emphasized - it is their code, after all, AI cannot be put responsible for anything.

jromanowska:
I have updated the text now, thanks for the suggestion, @sneakers-the-rat! Is it OK to merge?

jromanowska:
Since the other people in this thread did not have objections, I will merge this. This can be changed when needed. :+1:

PR #1412, opened 2025-03-09
Make votes easier to read

arfon:
Based on editor feedback about the short message being hard to read. Now we make it all visible!


arfon:
Looks good in prod.

PR #1415, opened 2025-04-02
Ask invited rev.s to reply even if not interested to speed up reviewer search.

fabian-s:
Ask invitees to reply even if not interested to speed up reviewer search.


arfon:
Good idea!

PR #1538, opened 2026-05-02
Security fixes

arfon:
Thanks to @andrew for alerting us to these and supplying the fixes 💖


app/models/paper.rb
```diff
@@ -521,7 +521,8 @@ def markdown_code
 private
 
   def check_repository_address
-    stdout_str, stderr_str, status = Open3.capture3("git ls-remote #{repository_url}")
```
sneakers-the-rat, line 524: Oh oop nice catch

PR #1543, opened 2026-05-07
Editor badges :)

sneakers-the-rat:
i had a few minutes free and was like hey why don't we have editor badges.

Pretty much the same thing as #1389 , which was written with this in mind anyway. hopefully self explanatory.

@arfon i don't know if we need the migration/fix from #1509 , since editors actually do have their own table/model? i see that there is some handling of the editor in the metadata dict though - unsure how to test rly because it works fine with the seed data, and not sure how to simulate the process that happens in prod so i can write an rspec test for it.

anyway, ezpz hopefully.

i made it purple because the footer is purple and the buttons are purple, but i lightened it up a tiny bit because in such a small area it was just reading as black. but obvi i can change it to match and also make it whatever other color

I removed the `visible` scope from this badge, since papers that were rejected or withdrawn should still count as performing your editorial duty!

Screenshot 2026-05-07 at 4 10 12 PM

### DIFF
--- a/app/controllers/badges_controller.rb
+++ b/app/controllers/badges_controller.rb
@@ -24,6 +24,18 @@ class BadgesController < ApplicationController
     }
   end
 
+  def edited_by
+    n_edits = Paper.visible.where(":editor = editor", editor: params[:editor]).count
+    @key = string_item("JOSS Editor", COLORS[:gray])
+    @value = string_item(n_edits.to_s, COLORS[:blue], @key[:outer_width])
+    @badge_width = @key[:outer_width] + @value[:outer_width]
+
+    render template: 'badges/badge', locals: {
+      scale_up_factor: SCALE_UP_FACTOR,
+      scale_down_factor: SCALE_DOWN_FACTOR
+    }
+  end
+
   private
 
   def string_item(str, color, offset=0)
--- a/app/controllers/badges_controller.rb
+++ b/app/controllers/badges_controller.rb
@@ -9,7 +9,8 @@ class BadgesController < ApplicationController
 
   COLORS = {
     gray: "#555",
-    blue: "#007ec6"
+    blue: "#007ec6",
+    purple: "#4f4686"
   }
 
   def reviewed_by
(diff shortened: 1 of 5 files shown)


app/controllers/badges_controller.rb
```diff
@@ -24,6 +25,20 @@ def reviewed_by
     }
   end
 
+  def edited_by
+    n_edits = Paper.joins("JOIN editors ON editors.id = papers.editor_id")
+                   .where(editors: { login: params[:editor] })
+                   .count
```
arfon: Given the weirdness sometimes of GitHub handles and CaSe sensitivity, I think we should lower these:

```suggestion
  login = params[:editor].to_s.delete_prefix('@').downcase
  n_edits = Paper.joins(:editor).where("lower(editors.login) = ?", login).count
```

sneakers-the-rat: you know it better than me, weird how github names are not case sensitive when i'm just used to thinking of "treat everything case sensitive by default"

sneakers-the-rat:
ut oh maybe need a linter or formatter to enforce this https://github.com/openjournals/joss/pull/1543/commits/391db03c187f9e6b4d19876f22e89836742dd48c

arfon:
Sorry, didn't realize you'd be working on this at this time too @sneakers-the-rat! Just pushed a few specs too to cover the new behavior.

spec/controllers/badges_controller_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+
+RSpec.describe BadgesController, type: :controller do
+  render_views
+
+  describe "GET #edited_by" do
+    let(:editor) { create(:editor, login: "editor-jane") }
+
+    # Pulls the count out of the SVG's aria-label, which is rendered as
+    # `aria-label="JOSS Editor: 3"` by app/views/badges/badge.svg.erb.
+    def rendered_count
```
sneakers-the-rat, line 11: oh nice this is a way tidier way of doing it than what i had drafted. i have come to really appreciate ruby and rspec even if i still find them perplexing sometimes, and this is one time when i really get the reason for ruby's implicitness. makes nice interface possible

arfon:
BTW, I've changed the target of the PR to a branch on this fork so I can deploy to Heroku

spec/controllers/badges_controller_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+      expect(rendered_count).to eq("0")
+    end
+
+    it "does not count papers edited by other editors" do
+      other = create(:editor, login: "someone-else")
+      create_list(:accepted_paper, 2, editor: other)
+      create(:accepted_paper, editor: editor)
+      get :edited_by, params: { editor: editor.login }, format: :svg
+
+      expect(rendered_count).to eq("1")
```
sneakers-the-rat, line 51: wait shouldn't this be 0

edit: nvm, there are two papers

sneakers-the-rat:
> Sorry, didn't realize you'd be working on this at this time too

insomnia is the gift i give to to the world

> I've changed the target of the PR to a branch on this fork so I can deploy to Heroku

[do what cha gotta do](https://www.youtube.com/watch?v=ZJUBxaL8PD4)

PR #1544, opened 2026-05-09
Editor badges (the production deploy)

arfon:
Screenshot 2026-05-09 at 10 43 06

Looking good in production, although the fact we don't scope to `visible` means that the count in the top right, is different from the paginated count... 

cc @sneakers-the-rat

### DIFF
--- a/app/controllers/badges_controller.rb
+++ b/app/controllers/badges_controller.rb
@@ -24,6 +24,18 @@ class BadgesController < ApplicationController
     }
   end
 
+  def edited_by
+    n_edits = Paper.visible.where(":editor = editor", editor: params[:editor]).count
+    @key = string_item("JOSS Editor", COLORS[:gray])
+    @value = string_item(n_edits.to_s, COLORS[:blue], @key[:outer_width])
+    @badge_width = @key[:outer_width] + @value[:outer_width]
+
+    render template: 'badges/badge', locals: {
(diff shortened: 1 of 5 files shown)


sneakers-the-rat:
>  the count in the top right, is different from the paginated count...

i think this is correct, if someone spends their entire life correctly rejecting junk papers (inshallah we never receive them) they should receive credit for that work. if someone notices the discrepancy and really cares, we can add a ？tooltip or something. 

coming from thinking about information leaks and fediverse dev stuff, feels obligatory to say "i don't think that disclosing the number of papers that have been rejected/etc. is an information leak that compromises protected participant information"

arfon:
> i think this is correct, if someone spends their entire life correctly rejecting junk papers (inshallah we never receive them) they should receive credit for that work. if someone notices the discrepancy and really cares, we can add a ？tooltip or something.

Yeah, fair enough. Just noting. I don't think we need to explain.

> coming from thinking about information leaks and fediverse dev stuff, feels obligatory to say "i don't think that disclosing the number of papers that have been rejected/etc. is an information leak that compromises protected participant information"

:+1:

matthewfeickert:
Thanks for doing this @sneakers-the-rat and @arfon. I think this is a nice piece of shininess :sparkles: and functionality that people will appreciate. :bow:

PR #1545, opened 2026-05-13
fix disabling browser autcompletion from history

aycsi:
## What
Disable browser autocomplete/past searches on the papers search input.

## Why
Browser history suggestions interferes and make search frustrating for users (e.g., when searching by author or tag and other things are popping up).

### DIFF
--- a/app/views/papers/index.html.erb
+++ b/app/views/papers/index.html.erb
@@ -32,10 +32,10 @@
       <% end %>
     </div>
 
-    <%= form_for("", url: search_papers_path, method: 'get', html: { class: "form-inline my-2 my-lg-0 float-right"}) do |f| %>
+    <%= form_for("", url: search_papers_path, method: 'get', html: { class: "form-inline my-2 my-lg-0 float-right", autocomplete: "off"}) do |f| %>
     <div class="input-group mb-3">
       <%- placeholder_text = params[:q].nil? ? "Search by title, tag, author, or language":params[:q]%>
-      <%= f.text_field :q, placeholder:placeholder_text , class: "form-control", size: "35" %>
+      <%= f.text_field :q, placeholder:placeholder_text , class: "form-control", size: "35", autocomplete: "off" %>
       <div class="input-group-append">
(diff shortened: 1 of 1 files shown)


sneakers-the-rat:
Could we get a screenshot maybe of the problem? Like an example of how auto complete is frustrating? I don't have strong reason to believe auto complete is helpful, I just want to see what the problem is

aycsi:
> Could we get a screenshot maybe of the problem? Like an example of how auto complete is frustrating? I don't have strong reason to believe auto complete is helpful, I just want to see what the problem is

It's basically "saved info" autofill from the browser. It can certainly be a non issue but I find it annoying. 

autocomplete="off" is a standard way to suppress browser history/autofill suggestions in that field.

sneakers-the-rat:
Oh like saved info from other sites, not prior searches in this field?

> autocomplete="off" is a standard way to suppress browser history/autofill suggestions in that field.

Yes, am aware, no dispute on the method, just looking to document the problem that is being solved.

aycsi:
> Oh like saved info from other sites, not prior searches in this field?
> 
> > autocomplete="off" is a standard way to suppress browser history/autofill suggestions in that field.
> 
> Yes, am aware, no dispute on the method, just looking to document the problem that is being solved.

Yeah :) 

I did a deeper dive into this, it seems like it differs between browsers. I've given it a go on non-cached browser states on different browsers and issue stays on Edge.

Can you test this on a fresh browser session as well?

sneakers-the-rat (approved):
can confirm that i am just getting autocomplete values from any search bar in my browser's history. Could be possible to make that autofill prior searches for just that field by giving it a name or id, but no autofill is probably better than "autofill from anything i have ever typed in an input field"

aycsi:
> can confirm that i am just getting autocomplete values from any search bar in my browser's history. Could be possible to make that autofill prior searches for just that field by giving it a name or id, but no autofill is probably better than "autofill from anything i have ever typed in an input field"

thanks for looking into it, I agree with you.

danielskatz:
👋 @arfon - Do you think we (you) can merge this?

PR #1551, opened 2026-06-12
Check author when showing scope notes and votes

### LINKED ISSUE #1550: Hide editor panels in paper view from editors who are the ones who made the submission
I just made my first submission to JOSS, and I can see the scope summary panel and the editor notes panels on my own submission. I probably should not be able to see these, so that other editors can speak candidly about my work and there is no special treatment.

I think all that would take is to add an additional condition here: https://github.com/openjournals/joss/blob/1cdbd4e1040ab4434da676d085ca61ab4d1c0a6d/app/views/papers/_show_unpublished.html.erb#L17-L18

to be like 

```erb
        <%= render partial: "papers/vote_summary", locals: { paper: @paper } if current_editor && not_author %>
        <%= render partial: "papers/notes", locals: { paper: @paper } if current_editor && not_author %>
```

and then have some kind of `not_author` (or `!is_author`, i always have to look up ruby syntax so idk if that's right) check to see if the currently logged in user is the submitting author (i don't think we store gh handles for other authors in the paper?)

this seems high-ish priority because it compromises editorial process unfairly

arfon:
This pull request introduces a new method to the `Paper` model to check if a given user is the submitting author and updates the UI to restrict editors from seeing certain sections for their own submissions. It also adds tests to ensure the new method works as expected.

**Model changes:**

* Added the `submitted_by?(user)` method to the `Paper` model to determine if a given user is the submitting author.

**UI changes:**

* Updated the unpublished paper view (`_show_unpublished.html.erb`) so that editors do not see the "vote summary" and "notes" sections for papers they themselves submitted.

**Testing:**

* Added tests for the `submitted_by?` method in `paper_spec.rb` to verify correct behavior for the submitting author, other users, and nil cases.

Fixes #1550

### DIFF
--- a/app/models/paper.rb
+++ b/app/models/paper.rb
@@ -211,6 +211,10 @@ class Paper < ApplicationRecord
     retraction_for_id.present?
   end
 
+  def submitted_by?(user)
+    user.present? && user_id == user.id
+  end
+
   def invite_editor(editor_handle)
     return false unless editor = Editor.find_by_login(editor_handle)
     Notifications.editor_invite_email(self, editor).deliver_now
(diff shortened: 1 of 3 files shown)


sneakers-the-rat (approved):
Yep this hides the panels on the deployed preview!

PR #1557, opened 2026-07-02
fix: Derive AEiC track display from track_aeics instead of free-text title

matthewfeickert:
## Description

On the about page's editorial board section, board editors rendered only the free-text title column, so an AEiC whose title was blank (never set, or wiped by the clear_title callback) showed no role or track at all.

Derive the label from the track_aeics association instead: board editors with managed tracks now always display "Associate Editor-in-Chief:" plus all of their managed tracks, falling back to the title column otherwise (e.g. for the Editor-in-Chief, or journals without the tracks feature).

I don't really know Ruby, so if there is a much better way to do this I would be happy to implement that instead.

## Motivation

I noticed that on the Editorial Board page my associated track doesn't render.

### View from 441c426ec938a41244ce17adf4bc2632e66f1907

image

## Local preview view from this PR

image

## Disclosure

The changes in this PR were generated by agentic model Claude Code Opus 4.8 with steering and input from me.

### DIFF
--- a/app/models/editor.rb
+++ b/app/models/editor.rb
@@ -93,6 +93,12 @@ class Editor < ApplicationRecord
     kind == "board"
   end
 
+  def board_title
+    return title unless JournalFeatures.tracks? && managed_tracks.any?
+
+    "Associate Editor-in-Chief: #{managed_tracks.map(&:name).join('; ')}"
+  end
+
   def board_removed?
     kind_changed? && kind_was == "board"
   end
--- a/app/models/editor.rb
+++ b/app/models/editor.rb
@@ -96,7 +96,11 @@ class Editor < ApplicationRecord
   def board_title
     return title unless JournalFeatures.tracks? && managed_tracks.any?
 
-    "Associate Editor-in-Chief: #{managed_tracks.map(&:name).join('; ')}"
+    if board? && title.blank? && JournalFeatures.tracks? && managed_tracks.any?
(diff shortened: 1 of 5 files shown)


matthewfeickert:
Tagging @xuanxu for review given the [history of `spec/models/editor_spec.rb`](https://github.com/openjournals/joss/commits/main/spec/models/editor_spec.rb).

app/models/editor.rb
```diff
@@ -93,6 +93,12 @@ def board?
     kind == "board"
   end
 
+  def board_title
+    return title unless JournalFeatures.tracks? && managed_tracks.any?
+
+    "Associate Editor-in-Chief: #{managed_tracks.map(&:name).join('; ')}"
```
xuanxu: I would add preference to the custom title if it exists, to cover cases like the EiC. Also added to the suggested change to check the editor is on the board:

```suggestion
    if board? && title.blank? && JournalFeatures.tracks? && managed_tracks.any?
      "Associate Editor-in-Chief: #{managed_tracks.map(&:name).join('; ')}"
    else
      title
    end
```

xuanxu (approved):
Other than the suggestion it looks good to me.

Pinging @arfon for a AEiC approval 😁

arfon:
Looks good in production

matthewfeickert:
> Looks good in production

Yup. :+1: https://joss.theoj.org/about now shows

image

Thanks for very speedy reviews!

PR #1560, opened 2026-07-07
Add note about removing optional fields from example paper

jatkinson1000:
I had a recent submission where the authors had copied the example paper from the docs and left the AAS tags in.

This PR adds a note to explicitly say these should be removed by authors if they are not relevant to the submission.

Alternatively, I would advocate for removing the optional flags entirely as the majority of submissions to JOSS (that will use this as a starting point) will not require them.
If they are part of a joint submission they can be made aware of the need to add a tag elsewhere (e.g. as was done for RSECon26). I leave this as an option for the reviewers/maintainers to discuss.

I also added reference to the fact that there may be other tags/joint submissions besides AAS, but am happy to leave it as just referencing AAS if preferred.

### DIFF
--- a/docs/example_paper.md
+++ b/docs/example_paper.md
@@ -40,7 +40,7 @@ date: 13 August 2017
 bibliography: paper.bib
 
 # Optional fields for papers that are part of a joint submission.
-# For example submitting to a AAS journal too, see this blog post:
+# For example, submitting to a AAS journal too, see this blog post:
 # https://blog.joss.theoj.org/2018/12/a-new-collaboration-with-aas-publishing
 #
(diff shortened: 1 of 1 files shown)


danielskatz:
I would remove the RSECon26 part, as I don't think we will get more submissions for it at this point, while the AAS part is ongoing. I do like the added line about removing the AAS part if not needed, however.

danielskatz (changes requested):
see comment in conversation about changes

jatkinson1000:
@danielskatz I have done that now.

danielskatz (approved):

xuanxu (approved):
:+1:

PR #1563, opened 2026-07-19
Automated / assisted scope-review pipeline

arfon:
> **This is an experiment we're running with the editorial team, not a finished feature.** It ships turned off, gets switched on gradually, and nothing it produces is binding — editors make every decision. The whole point of the rollout is to find out whether it's genuinely helpful and trustworthy before anyone relies on it, and to change or drop it based on what editors tell us.

## What this adds

Every new JOSS submission now needs an editorial scope check before an editor is assigned — enough public development history, real research impact, open-source practice including tests, community context, and steady development over time. Today an editor does this by hand for every paper, which is slow and mostly repetitive.

This PR adds a system that does the repetitive part automatically and drafts a recommendation, so an editor can clear the obvious cases quickly and spend their time on the genuinely tricky ones. It is an assistant, not an autopilot: it reads the paper and the repository and writes up a suggestion, but a human makes and sends every actual decision.

It ships **turned off** in production (`scope_review.enabled: false`) so it can run quietly alongside real editorial decisions first, and we can check it agrees before anyone relies on it.

Two documents are the best place to start:

- [`docs/scope_review_for_editors.md`](https://github.com/openjournals/joss/blob/scope-review-pipeline/docs/scope_review_for_editors.md) — written for editors: what you see, what you do, and how to change how it behaves later (including editing the AI's instructions, which needs no code).
- [`docs/scope_review.md`](https://github.com/openjournals/joss/blob/scope-review-pipeline/docs/scope_review.md) — the technical version: how it's built, how to deploy it, and the rollout plan.

## The ideas it's built on

A few principles shaped the whole design, and they're the useful lens for reviewing it:

- **The AI drafts; it never decides or acts.** The things that can be checked as facts — repository age, whether tests exist, the licence — are computed directly, not asked of the AI. The AI's job is to read the paper and suggest a recommendation for a human to consider. It can weigh in on things facts alone can't settle, but it can never overturn a computed fact, and nothing it produces sends an email or touches GitHub.
- **"Don't know" is never treated as "fail."** When something can't be decided automatically — missing data, an unusual code host — the paper is handed to a human to look at, never rejected on its own.
- **If anything goes wrong, it asks for help instead of guessing.** Any error, timeout, or safety limit sends the paper to the manual queue rather than producing a verdict.
- **It won't judge a paper it can't read.** If there's no `paper.md`, it doesn't assess — it asks the author to add the paper.
- **It's hard to manipulate.** A submitter can't hide instructions in their paper to force a decision: the paper is treated as untrusted text, and the deeper AI investigation can only *read* a fixed set of GitHub pages for the submission's own repository — it can't be pointed elsewhere or made to change anything.

## How the review is organised

It works in stages, cheapest first, and only goes deeper when it needs to. Most papers are settled in the early, free stages; only the genuinely ambiguous ones reach the slow, more expensive AI investigation. Here's a walkthrough of the code in a sensible reading order.

**Start with the conductor.** [`Runner#assess`](https://github.com/openjournals/joss/blob/scope-review-pipeline/app/services/scope_review/runner.rb#L21-L48) runs the whole thing end to end and is where all the "ask a human instead of guessing" safeguards live — an oversized repository, a failed clone, or any error all send the paper to the manual queue rather than to a verdict. [`assess_checkout`](https://github.com/openjournals/joss/blob/scope-review-pipeline/app/services/scope_review/runner.rb#L50-L77) is where it decides how deep to go, and where it refuses to continue if there's [no paper to read](https://github.com/openjournals/joss/blob/scope-review-pipeline/app/services/scope_review/runner.rb#L79-L88).

Two spots in the conductor are worth a careful look because they enforce the "AI never overrules the facts" principle:

- [`merge_model_gates`](https://github.com/openjournals/joss/blob/scope-review-pipeline/app/services/scope_review/runner.rb#L185-L194) — the AI's opinion can only fill in a check the facts left undecided; it can never flip a computed result.
- [`cap_impact_only_reject`](https://github.com/openjournals/joss/blob/scope-review-pipeline/app/services/scope_review/runner.rb#L150-L162) — if a paper is strong on everything except demonstrated research impact, the system will *not* reject it on its own; it hands that judgement call to an editor. This came directly out of testing against real decisions.

**The fact-based checks** live in [`Gates`](https://github.com/openjournals/joss/blob/scope-review-pipeline/app/services/scope_review/gates.rb#L56-L73). Each check comes out as pass, fail, or "not sure", and [anything unsure sends the paper to a human or the deeper AI look](https://github.com/openjournals/joss/blob/scope-review-pipeline/app/services/scope_review/gates.rb#L278-L290) rather than failing it.

**The deeper AI investigation** is the security-sensitive part, so it's kept on a tight leash. It can use exactly one tool — reading a [fixed list of GitHub pages](https://github.com/openjournals/joss/blob/scope-review-pipeline/app/services/scope_review/investigator.rb#L25-L28) for the submission's own repository ([and nothing else](https://github.com/openjournals/joss/blob/scope-review-pipeline/app/services/scope_review/investigator.rb#L244-L255)) — under [strict time and step limits](https://github.com/openjournals/joss/blob/scope-review-pipeline/app/services/scope_review/investigator.rb#L14-L19), and the paper is [handed to it clearly marked as untrusted](https://github.com/openjournals/joss/blob/scope-review-pipeline/app/services/scope_review/prompts.rb#L49-L64).

**The facts it gathers** come from the repository itself, using plain `git` rather than heavier tooling so there's nothing new to install on Heroku. A couple of pieces are worth sanity-checking: how it [detects a repository "dumped" all at once versus genuinely developed over time](https://github.com/openjournals/joss/blob/scope-review-pipeline/app/services/scope_review/git_history.rb#L80-L98), and how it [ignores bot accounts](https://github.com/openjournals/joss/blob/scope-review-pipeline/app/services/scope_review/git_history.rb#L100-L108) when counting contributors (a real bug caught in testing — a solo project plus dependabot was briefly counted as collaborative).

**The AI's actual instructions** are in [`config/scope_review/instructions.md`](https://github.com/openjournals/joss/blob/scope-review-pipeline/config/scope_review/instructions.md) — plain prose, given to the AI word for word. This is the main dial for changing how it judges papers, and editing it needs no programming. The one rule: the fact-based checks in code and this file describe the same criteria, so if you change a criterion that has a mechanical part (like the six-month rule) both have to change together.

**What an editor sees and does** is the [assessment card on the paper page](https://github.com/openjournals/joss/blob/scope-review-pipeline/app/views/papers/_scope_assessment.html.erb) and a screening column on the incoming dashboard. The one place an outward action can happen is [`desk_reject`](https://github.com/openjournals/joss/blob/scope-review-pipeline/app/controllers/papers_controller.rb#L238-L263), which rejects a paper and optionally emails the author the (editable) draft — always behind a human click.

Assessments are stored per paper and kept forever (a [new record each time a paper is re-checked](https://github.com/openjournals/joss/blob/scope-review-pipeline/app/models/scope_assessment.rb)), so there's a full audit trail. There are three [rake tasks](https://github.com/openjournals/joss/blob/scope-review-pipeline/lib/tasks/scope_review.rake): one to screen the whole existing backlog once at deploy, one for the ongoing scheduled sweep, and one to re-check a single paper. Tests are in [`spec/services/scope_review/`](https://github.com/openjournals/joss/tree/scope-review-pipeline/spec/services/scope_review).

## Deploying it

- One new setting is needed: **`ANTHROPIC_API_KEY`**. The GitHub token is reused (already set), and there are no changes to the Heroku build or stack.
- All the tunable settings — which AI models, thresholds, size limits, and the on/off switch — live in `config/settings-production.yml`, not in environment variables.
- The sequence: deploy → set `ANTHROPIC_API_KEY` → run `rake scope_review:backlog` once to screen the existing backlog → set `enabled: true` and add the scheduled job.

## How well it works

Tested against 40 real editorial decisions (using a copy of the production database): on 20 papers that were rejected, it agreed with 17 and correctly sent the rest to manual review. On 20 papers that passed the scope check, after adding the "don't reject on impact alone" safeguard, it produced **no false rejections**. Where it does err, it errs toward asking a human rather than rejecting on its own.

## Screenshots
Screenshot_2026-07-19_at_22_18_29

Screenshot_2026-07-19_at_22_19_02

### DIFF
--- a/docs/scope_review_for_editors.md
+++ b/docs/scope_review_for_editors.md
@@ -15,7 +15,7 @@ the system architecture see [`scope_review.md`](scope_review.md).
 
 ## What it is for
 
-Since 2026 every new submission needs an editorial scope check before an editor
+Since 2026 new submissions needs additional editorial checks before an editor
 is assigned: does it show enough public development history, demonstrated
 research impact, open-source practice (including automated tests), community
 context, and iterative development? Doing this by hand for every submission is
--- a/docs/scope_review_for_editors.md
+++ b/docs/scope_review_for_editors.md
@@ -1,5 +1,13 @@
 # Automated scope screening — a guide for editors
 
+> **This is an experiment we are running with the editorial team.** It is new,
+> switched on deliberately and gradually, and it is not a settled part of the
+> workflow. Nothing it produces is binding — editors make every decision, as
+> always. We are running it precisely to find out whether it is actually
+> helpful and trustworthy, and your feedback (and your agree/override clicks —
+> see below) is how we decide whether to keep it, change it, or drop it. Please
+> tell us what's wrong with its suggestions; that's the point of the exercise.
+
 This explains what the automated scope screening is for, what you see and do
 as an editor, and — for whoever maintains JOSS — how to change how it behaves
 later. It deliberately avoids the mechanical detail of each check; for that and
--- a/docs/scope_review_for_editors.md
+++ b/docs/scope_review_for_editors.md
@@ -1,0 +1,179 @@
+# Automated scope screening — a guide for editors
+
+This explains what the automated scope screening is for, what you see and do
+as an editor, and — for whoever maintains JOSS — how to change how it behaves
+later. It deliberately avoids the mechanical detail of each check; for that and
+the system architecture see [`scope_review.md`](scope_review.md).
+
+## What it is for
+
+Since 2026 every new submission needs an editorial scope check before an editor
+is assigned: does it show enough public development history, demonstrated
+research impact, open-source practice (including automated tests), community
+context, and iterative development? Doing this by hand for every submission is
+slow and repetitive, even though most of the signal is mechanical.
+
+The screening automates the **mechanical triage and the first draft of the
+decision**, and stops there. It is an assistant, not an autopilot:
+
+- It reads text and metadata — the paper, the git history, the repository tree,
+  the licence, GitHub activity. It never runs the submitted code.
+- It produces a **recommendation and a draft note**, never an action. No email
+  is sent and no GitHub issue is touched unless *you* click to do it.
+- Its goal is to save you time on the clear cases and to gather the evidence for
+  the genuinely ambiguous ones — not to make the call for you.
+
+The irreversible, outward-facing decision always stays with a human.
+
+## What you see
+
+Two places surface the screening, both editor-facing:
+
+- **The incoming dashboard** ("Papers with no editor") has a **Screening**
+  column showing each paper's recommendation as a small badge. This is for
+  triage at a glance — spotting the clear desk-rejects in a long queue.
+- **The paper page** carries an **assessment card**: the recommendation, each
+  criterion with a plain-language result and a hover explanation, any anomalies
+  the automation flagged, an editable draft note, and the decision buttons.
+  There is also a queue of everything awaiting review at `/scope_assessments`.
+
+### The recommendations
+
+| Recommendation | What it means for you |
+|---|---|
+| **Proceed** | Looks clear on every criterion. Assign an editor as usual. |
+| **Borderline proceed** | Leans proceed, with a caveat worth a glance. |
+| **Requires verification** | Genuinely ambiguous — the automation gathered evidence but wants a human to make the call. Read the summary. |
+| **Desk reject** | A clear failure (usually a mechanical one like too little history or no tests). Review the draft note and decide whether to send it. |
+| **Needs manual** | The automation couldn't assess it at all (no paper file found, repository too large to clone, an error). Assess by hand. |
+
+Two things to internalise, because they are easy to misread:
+
+- A criterion marked **"?" (unknown) never counts against a submission.** It
+  means "not decided automatically", not "failed". Research impact, for
+  instance, is *always* left to judgment — it is never failed by the machine
+  alone.
+- A recommendation is **advisory**. Recording your decision is what carries
+  editorial weight, and the buttons below make that explicit.
+
+## What you do
+
+On the assessment card:
+
+- **Desk reject** rejects the paper and, if you tick the box, emails the
+  (editable) draft note to the submitting author. This is for clear failures
+  that never need a public review issue — it keeps the process quieter and
+  gentler for the author. Nothing is sent until you click, and you can edit the
+  note first.
+- **Agree** / **Override** record whether the automation got it right, *without*
+  touching the paper. Use these when you're actioning the outcome elsewhere
+  (e.g. on GitHub) but still want to log whether the screening was correct.
+- **Re-run** re-screens the paper — useful after an author pushes a fix (adds a
+  paper file, adds tests).
+
+**Please record a decision even when you act elsewhere.** Every agree/override
+is calibration data: it is how we measure whether the automation is trustworthy
+and where it needs tuning. During the initial rollout the screening runs
+silently alongside your normal decisions precisely so we can compare the two
+before relying on it.
+
+## Why some assessments are instant and some take minutes
+
+The screening works in tiers, cheapest first, and only escalates when it needs
+to:
+
+1. **The clear cases** are decided from repository facts alone — no AI model, a
+   few seconds, effectively free. A repository that is three months old or has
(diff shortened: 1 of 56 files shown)


docs/scope_review_for_editors.md
```diff
@@ -0,0 +1,187 @@
...
+> tell us what's wrong with its suggestions; that's the point of the exercise.
+
+This explains what the automated scope screening is for, what you see and do
+as an editor, and — for whoever maintains JOSS — how to change how it behaves
+later. It deliberately avoids the mechanical detail of each check; for that and
+the system architecture see [`scope_review.md`](scope_review.md).
+
+## What it is for
+
+Since 2026 every new submission needs an editorial scope check before an editor
```
danielskatz: This has always been the case - in 2026 what that scope check is changed.

danielskatz:
I feel like this is something that we should have a discussion about.

I'm particularly concerned about the recommendation parts, not the data gathering parts.

arfon:
> I feel like this is something that we should have a discussion about.

Definitely, playing around at this point (I meant to mark this as draft).

docs/scope_review_for_editors.md
arfon:
```suggestion
Since 2026 new submissions needs additional editorial checks before an editor
```

sneakers-the-rat:
a machine should never make an editorial decision. a machine should not pollute the editorial decision with stochastic prose evaluations. consolidating the deterministic checks into a report seems like a good move, but under no circumstances should we be asking an LLM to triage papers - even if the LLM can only decide to escalate to human review, it frames the decision that the human editor makes rather than the submission itself.

jedbrown:
I agree with @sneakers-the-rat. Automation bias is strong in this context and LLMs embody biases that the editor may not even be aware of. "Accuracy" in historical decisions does not imply validity. If we were to deploy this and then audit the resulting system, we would almost certainly find a lack of construct validity and that "bad" recommendations from the LLM unacceptably influenced decisions. I think it also violates trust of authors and the broader community, and cheapens the meaning of a JOSS publication.

There are also possible legal concerns in some jurisdictions since paper rejections may be "consequential decisions" and thus subject to a variety of disclosure/appeal/audit requirements. I doubt any of the current legislation or rulemaking would exclude this proposed use, but that may not hold. Whether First Amendment protection applies (to enable possibly-discriminatory or "invalid" editorial decisions) in the US probably depends on whether JOSS publication is considered to be _transactional_ (e.g., more like a paid columnist versus an unpaid op-ed at a newspaper). In any case, I have been doing legislative/policy work around algorithmic decision systems and would be horrified for (perceived or effective, due to automation bias) JOSS robo-scope-review to be held up to undercut my arguments about audits and duties of care.

arfon:
Good feedback, thanks @sneakers-the-rat, @jedbrown, @danielskatz – and this is definitely marked as "Draft" for a reason...

A couple of points to note:

- The decision is never automated here. Decisions in my opinion should always reside with humans. 
- The goal is to gather context (deterministic and interpreted) to assist the triage process. I understand concerns about the latter.

Track EiCs I believe need help here with scope reviews. In my opinion (as outgoing EiC), far too much time is spent [rejecting](https://www.theoj.org/joss-analytics/joss-submission-analytics.html#Fraction_rejected_papers) low effort vibe-coded submissions, which leaves precious little time to attend to the more nuanced editorial work such as dealing with editorial questions that come up during reviews, finding the right editor for a paper etc.

Lots to chew on here. Looking forward to discussing in depth and figuring out the right path forward.

kthyng:
It's taken me this long to get here to comment but I kept it unread so I could get here. I just want to say that, indeed, it is a lot of work to address the items that need to be checked and I am personally falling behind. I don't think the current way we're doing this is sustainable in terms of human time, or least in terms of my human time.

> Track EiCs I believe need help here with scope reviews. In my opinion (as outgoing EiC), far too much time is spent [rejecting](https://www.theoj.org/joss-analytics/joss-submission-analytics.html#Fraction_rejected_papers) low effort vibe-coded submissions, which leaves precious little time to attend to the more nuanced editorial work such as dealing with editorial questions that come up during reviews, finding the right editor for a paper etc.

Issue #1376, opened 2024-10-22
Problems with indexing under Google Scholar


MartinBeseda:
Hello everybody,
recently (~ 4 weeks ago) we've published a paper under JOSS (https://joss.theoj.org/papers/10.21105/joss.06036), while the pre-print was published on ArXiv before (https://arxiv.org/abs/2401.11884).

Today I found a puzzling thing, where the JOSS paper is not indexed by Google Scholar yet (while ArXiv is indexed in the matter of days), but it "returns" the ArXiv pre-print, when looking for JOSS DOI. 

![screen](https://github.com/user-attachments/assets/e80c8af1-65cc-42b0-922d-c194a7033eca)

I'm not sure, what's happening here, as GS seems to find the DOI correctly together with the paper's title, but it seems unable to index the JOSS page, so it resorts to returning only ArXiv version, which is visible. Now, as the crawler clearly found the title corresponding to DOI already, I find this a little disturbing.

 This issue may or may not be directly connected to problems mentioned in #130 .

arfon:
> I'm not sure, what's happening here, as GS seems to find the DOI correctly together with the paper's title, but it seems unable to index the JOSS page, so it resorts to returning only ArXiv version, which is visible. Now, as the crawler clearly found the title corresponding to DOI already, I find this a little disturbing.

I don't think we know that this is the case (that Google Scholar can't index the JOSS page). A quick check of a random paper from two weeks ago finds the JOSS record correctly: https://scholar.google.com/scholar?hl=en&as_sdt=0%2C5&q=%22Statmanager-kr%3A+A+User-friendly+Statistical+Package+for+Python+in+Pandas%22&btnG=

My guess is that there's something weird going on here with Scholar and the fact that the paper is also on arXiv. 

Other than ensuring that the correct metadata are present (which they are for your paper), unfortunately this team has no control over how Google Scholar behaves. 

![Image](https://github.com/user-attachments/assets/d253611b-30b8-45e5-b93e-84968212abbd)

As such, I'd encourage you to take this up with the Google Scholar team.

sneakers-the-rat:
I'm asking some scholcomm people if there's anything extra we need to do to indicate that the JOSS version is the version of record that google scholar would recognize, but looking at a few big journal sites they don't seem to do anything special. 

[...]

PR #1565, opened 2026-07-26
Making PDF links in paper#show consistent

arfon:
This pull request updates how the PDF download link for published papers is handled, ensuring that the download uses the SEO-friendly URL and that the PDF is proxied instead of redirected. It also updates the tests to match this new behavior.

**PDF Download Link Handling:**

* Updated the `seo_pdf_url` method in `paper.rb` to clarify that the PDF is now proxied via `PapersController#show` rather than being redirected, making the download URL consistent with the paper landing page.
* Changed the download button in `_sidebar_published.html.erb` to use `@paper.seo_pdf_url` instead of `@paper.pdf_url`, ensuring users download the PDF from the SEO-friendly URL.

**Test Adjustments:**

* Updated the system spec in `show_published_spec.rb` to expect the "Download paper" link to point to `@accepted_paper.seo_pdf_url`.

### DIFF
--- a/app/models/paper.rb
+++ b/app/models/paper.rb
@@ -358,8 +358,8 @@ class Paper < ApplicationRecord
   end
 
   # Return the seo_url plus '.pdf'. This is for Google Scholar so that we can
-  # point to PDFs on the same domain on the JOSS site (although they are then
-  # 301 redirected to a pdf_url later).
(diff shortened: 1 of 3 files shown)


arfon:
Possibly related to #1376

PR #1566, opened 2026-07-31
docs: add Podman instructions alongside Docker (fixes #1552)

### LINKED ISSUE #1552: Instructions for podman
The instructions on https://joss.readthedocs.io/en/latest/paper.html#docker are specific to Docker. Podman is an alternative on many linux platforms that does not require root. The adjusted command is

```
 podman run --rm \
    --volume $PWD/paper:/data \
    --userns=keep-id \
    --env JOURNAL=joss \
    docker.io/openjournals/inara
```

It would be helpful to have it in the official documentation.

casper06:
## Summary

Adds Podman instructions alongside the existing Docker instructions for
compiling a paper draft locally.

Podman is a rootless alternative to Docker available on many Linux
platforms. The added section uses `--userns=keep-id` (Podman's rootless
equivalent to matching file ownership between host and container) instead
of Docker's `--user $(id -u):$(id -g)`.

Fixes #1552

## Changes

- `docs/paper.md`: new "Podman" subsection right after the existing
  "Docker" subsection, same style/formatting.

## Notes

The exact command was proposed by @MarDiehl in the issue.

### DIFF
--- a/docs/paper.md
+++ b/docs/paper.md
@@ -393,6 +393,18 @@ docker run --rm \
     openjournals/inara
 ```
 
+### Podman
+
+If you use [Podman](https://podman.io) instead of Docker, you can compile a draft of your paper locally with the same container image. Podman does not require root privileges and is available on many Linux distributions:
+
+```text
+podman run --rm \
+    --volume $PWD/paper:/data \
+    --userns=keep-id \
+    --env JOURNAL=joss \
+    docker.io/openjournals/inara
+```
+
 ### Locally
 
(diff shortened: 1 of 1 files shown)


MarDiehl (changes requested):
thanks!. I'd propose the following change:

"Podman does not require root privileges on many Linux platforms:" should be "Podman does not require root privileges and is available on many Linux distributions:"

casper06:
Updated the wording as suggested — thanks for the feedback!

PR #1567, opened 2026-07-31
fix: disable misspellings/fuzzy matching for tag search (fixes #1536)

### LINKED ISSUE #1536: tagged papers list should not use edit-distance similarity checks for short tags
https://joss.theoj.org/papers/tagged/CWL/ lists 7 papers, but only one of those lists CWL in its tags or mentions `CWL` or `Common Workflow Language`.

I suggest that for short tag searches, they not include spelling variations, except for maybe allowing plurality suffixes or similar

2026-04-28 snapshot of the list of entries from https://joss.theoj.org/papers/tagged/CWL/ 

1. https://joss.theoj.org/papers/10.21105/joss.00023 ; OWL → CWL ?
2. https://joss.theoj.org/papers/10.21105/joss.00087 ; CTL → CWL ?
3. https://joss.theoj.org/papers/10.21105/joss.00230; CRU → CWL ?
4. https://joss.theoj.org/papers/10.21105/joss.00900 ; CSL → CWL ?
5. https://joss.theoj.org/papers/10.21105/joss.01812 ; OWL → CWL ?
6. https://joss.theoj.org/papers/10.21105/joss.02879 ; BEM → CWL ? 
7. https://joss.theoj.org/papers/10.21105/joss.02664 actually mentions CWL

casper06:
## Summary

Fixes false positives in the tag search (`/papers/tagged/:tag`) caused by
Searchkick's default fuzzy matching (misspellings) kicking in on short tags.

As reported in the issue, searching for `CWL` matched papers tagged `OWL`,
`CTL`, `CRU`, `CSL`, and `BEM` — all edit-distance 1 from "CWL" under
Searchkick's default settings. The `author`, `editor`, and `reviewer`
branches of the same `filter` action already disable this with
`misspellings: false`; the `tag` branch was missing it.

## Changes

- `app/controllers/papers_controller.rb`: add `misspellings: false` to the
  tag search branch, consistent with the other three branches.
- `spec/controllers/papers_controller_spec.rb`: add a regression test
  asserting `Paper.search` is called with `misspellings: false` when
  filtering by tag.

## Note on scope

The tag search also matches against `:title` (`fields: [:tags, :title]`).
This means a paper without the `CWL` tag but with "CWL" mentioned in its
title would still appear in `/papers/tagged/CWL/`. I left this untouched
since it's a separate behavioral question (should tag search only match
curated tags, or also full text?) that maintainers may want to decide
independently of this bug fix.

## Testing

I wasn't able to run the test suite locally (no Ruby/Elasticsearch
environment set up on this machine), so the new spec hasn't been executed
yet — happy to iterate based on CI results or review feedback.

Fixes #1536

### DIFF
--- a/app/controllers/papers_controller.rb
+++ b/app/controllers/papers_controller.rb
@@ -117,7 +117,7 @@ class PapersController < ApplicationController
       @term = "reviewed by #{params['reviewer']}"
       @badge = {type: "reviewer", value: params['reviewer']}
     elsif params['tag']
-      @papers = Paper.search(params['tag'], fields: [:tags, :title], order: { accepted_at: :desc },
+      @papers = Paper.search(params['tag'], fields: [:tags, :title], misspellings: false, order: { accepted_at: :desc },
                   page: params[:page],
                   per_page: 10)
       @term = "#{params['tag']}"
--- a/spec/controllers/papers_controller_spec.rb
+++ b/spec/controllers/papers_controller_spec.rb
@@ -610,6 +610,34 @@ describe PapersController, type: :controller do
     end
   end
 
+  describe "GET #filter by tag" do
+    # Stands in for a Searchkick::Results object: the controller passes it to
+    # Pagy.new_from_searchkick and the view iterates over it.
+    let(:empty_search_results) do
+      instance_double(Searchkick::Results,
+                      total_count: 0,
+                      options: { page: 1, per_page: 10 }).tap do |results|
+        allow(results).to receive(:each)
+        allow(results).to receive(:size).and_return(0)
+        allow(results).to receive(:to_ary).and_return([])
+      end
+    end
+
+    # Searchkick's default misspellings setting applies an edit distance of 1 to
+    # short terms, so a search for a tag like "CWL" also matched papers tagged
+    # "OWL", "CTL" or "CSL". See https://github.com/openjournals/joss/issues/1536
+    it "disables misspellings/fuzzy matching when filtering by tag" do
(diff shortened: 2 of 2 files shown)


app/controllers/papers_controller.rb
```diff
@@ -117,7 +117,7 @@ def filter
       @term = "reviewed by #{params['reviewer']}"
       @badge = {type: "reviewer", value: params['reviewer']}
     elsif params['tag']
-      @papers = Paper.search(params['tag'], fields: [:tags, :title], order: { accepted_at: :desc },
+      @papers = Paper.search(params['tag'], fields: [:tags, :title], misspellings: false, order: { accepted_at: :desc },
```
sneakers-the-rat, line 120: Preexisting, but unclear to me why tag search would include the title.

casper06:
Good catch — flagged the same thing in the PR description under "Note on
scope". I intentionally left `:title` out of this fix since it felt like
a separate design question (should tag search only match curated tags,
or also full text?) rather than part of the edit-distance bug this PR
addresses.

Happy to remove `:title` here too if you think it belongs in the same PR,
or I can open it as a follow-up issue/PR instead — whichever you prefer.

app/controllers/papers_controller.rb
arfon: LoL. I'm unclear either. But perhaps we should leave as-is?

arfon:
Verified in prod, with only the one match now for the paper _actually_ tagged with CWL: 

Screenshot 2026-08-26 at 17 49 45

PR #1571, opened 2026-08-25
Making language consistent with #1554

arfon:
This pull request makes minor clarifications to the "Scope and significance" section of the `about.html.erb` page. The changes refine the criteria for research impact and simplify the language for better clarity.

- Content clarifications:
  * Updated the description to specify that JOSS publishes articles about software that has demonstrated clear research impact, rather than just credible scholarly significance.
  * Simplified the list of evidence for research impact by removing "credible near-term significance demonstrated through benchmarks and reproducible materials" from the examples.

### DIFF
--- a/app/views/home/about.html.erb
+++ b/app/views/home/about.html.erb
@@ -35,12 +35,12 @@
 
         <h2>Scope and significance</h2>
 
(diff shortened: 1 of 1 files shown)


danielskatz:
thanks for picking this up and making it consistent with other recently changes docs - lgtm

arfon:
Thanks. Note this isn't deployed on the live site yet.

PR #1576, opened 2026-09-19
Adding issue activity overview to AEiC dashboard

arfon:
New admin view designed to track number of distinct issues/(review or pre-review) that people are commenting on. Designed to track un-solicited (and sometimes) spammy requests.


danielskatz:
Have you checked this? It seems like it it's showing some activity in the last 30 days that I would have expected.

I also wonder about the fact that the people who choose reviewers are editors, but I think this is just visible to AEiCs.

arfon:
It looks right to me? To be clear, there's no 'backfill' of old data here (as we don't have it), so this is a go-forward capability (i.e., will start tracking spurious activity going forward).

Screenshot 2026-09-19 at 16 16 41

Good thought about moving it out from the AEiC dashboard. It should be available to all editors. I'll work on that.

danielskatz:
Ok, no backfill explains what I saw as missing data.

arfon:
Sorry for not making that clear 😀

PR #1579, opened 2026-09-21
Sync reviewers from the GitHub issue header on webhook edits

arfon:
## Problem

editorialbot's `add/remove @x as reviewer` edits the `<!--reviewers-list-->` header in the issue body and notifies the OJRA reviewers service, but never calls back to JOSS (`api_assign_reviewers` has no caller anywhere in the org). The `papers.reviewers` column is therefore only set once, when the review issue is created via `api_start_review`, and drifts afterwards.

Measured against production today: **62 of 374 in-progress papers** differ from GitHub, and every paper still in pre-review has an empty reviewers column. This is why assigned reviewers were showing as "no role" on the new Issue activity tab (#1576, #1578).

## Fix

- New `IssueHeader` module parses the reviewers-list (and editor) markers from an issue body.
- `Paper#sync_reviewers_from_issue_body` updates the column when the parsed list differs (case-insensitive). Source of truth is the review issue once it exists, otherwise the pre-review issue.
- `DispatchHelper#process_edit` calls it on every `issues.edited` webhook. The payload always carries the full current body, verified against the recorded fixture and the live API.
- Legacy bodies without markers are left untouched.

## Backfill

`rake utils:sync_reviewers_from_github` (dry run by default, `DRY_RUN=false` to apply) walks `review_pending` and `under_review` papers, reads the header from GitHub, updates the column, and relabels that paper's `issue_comments` rows recorded as `none` for logins who are now reviewers.

## Test plan
- [x] Unit specs for `IssueHeader` parsing (markers absent, `Pending`, dedupe, non-handles)
- [x] Model specs for `sync_reviewers_from_issue_body` (add, remove, clear, no-op, legacy)
- [x] Request specs for the webhook: syncs on review-issue edit, ignores pre-review edits once a review issue exists, leaves legacy bodies alone
- [x] Full non-system suite: 413 examples, 0 failures
- [ ] After deploy: `heroku run rake utils:sync_reviewers_from_github -a joss-production`, review output, then rerun with `DRY_RUN=false`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


arfon:
@xuanxu – the alternative implementation here would be to add the ability for Editorialbot to post reviewer information when someone does "@editorialbot add XXX as reviewer" mid-review, but this seems equally good?

arfon:
Also, this was the issue that @danielskatz flagged [in Slack](https://openjournals.slack.com/archives/G016161K539/p1789906953953499).

spec/lib/issue_header_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+describe IssueHeader do
+  describe ".reviewers" do
+    it "returns nil when the body has no reviewers-list markers" do
+      expect(IssueHeader.reviewers("**Reviewer:** @mschubert , @JohnCoene")).to be_nil
+      expect(IssueHeader.reviewers(nil)).to be_nil
+    end
+
+    it "parses handles, normalizing the leading @ and dropping duplicates" do
+      body = "**Reviewers:** <!--reviewers-list-->@alice, bob ,@Alice, @carol-1<!--end-reviewers-list-->"
+      expect(IssueHeader.reviewers(body)).to eq(["@alice", "@bob", "@Alice", "@carol-1"])
```
xuanxu: Do we want here to return both `@Alice` and `@alice`?

xuanxu (commented):
Other than the small doubt about same user with different casing (an edge case that probably has not happened ever), this LGTM! :+1:

spec/lib/issue_header_spec.rb
arfon: Good catch, no. GitHub logins are case-insensitive so those are the same person. Fixed in a0a5d30: handles are now deduplicated case-insensitively, keeping the first spelling seen, and the spec asserts `["@alice", "@bob", "@carol-1"]` for that input.

xuanxu (approved):
Nice!

danielskatz:
Thanks @xuanxu!  Can you post a short note in out slack about this?

xuanxu:
@danielskatz Done!

PR #1580, opened 2026-09-21
Add a submission block list by ORCID iD, email, or repository

arfon:
## Summary

Lets AEiCs stop repeat spam submitters. A block list entry is an **ORCID iD**, an **email address**, or a **repository** (a single repo, or just an owner such as `github.com/owner`, which blocks every repository under that owner).

- **Enforcement**: a validation on `Paper` create. Blocked submitters see a deliberately generic message ("We're unable to accept this submission. If you believe this is an error, please contact the editorial team.") so they don't learn which detail matched. The `git ls-remote` check is skipped for blocked submissions. Existing papers are never affected.
- **Matching** is case-insensitive. Repository matching ignores scheme, `www.`, trailing slashes and `.git`. ORCID iDs can be entered as a bare iD or an orcid.org URL. Email entries are exact addresses only; bare domains such as `@gmail.com` are rejected, so a block can never affect other users at the same provider.
- **Management**: new "Block list" tab on the AEiC dashboard (`/blocklist`) to add and remove entries. Each entry records the editor who added it and a **required reason**.
- **Quick actions** on a paper's admin page: a **"Block all"** form that blocks the author's ORCID iD, email and repository owner together with one reason (idempotent, skips anything already blocked), plus individual buttons for each. The card shows what is already blocked.
- **Access**: `require_aeic` (board editors only), same as the rest of the AEiC dashboard. Topic editors are redirected.

Note on email: the profile email is never verified, so it is a backstop for catching a blocked ORCID holder who re-registers with the same address, not a primary control. The UI says so.

## Test plan
- [x] Model specs: normalization, ORCID/email format, bare domains rejected, owner-level repository matching, duplicate/kind/reason validation, `block_all_for` candidates/skip/blank handling
- [x] Paper specs: blocked ORCID, email, repo and owner; other addresses at the same domain allowed; update of existing paper unaffected; no `ls-remote` for blocked submissions
- [x] Controller specs: auth for non-AEiCs, index, create (normalization, attribution, required reason, domain refused, redirect back to referrer), block_paper (creates three entries, requires reason, idempotent, AEiC-only), destroy
- [x] Papers controller: blocked create returns 422 with the generic message and no mention of blocking; admin page shows "Block all" plus the individual buttons and the already-blocked state
- [x] Full non-system suite: 446 examples, 0 failures

🤖 Generated with [Claude Code](https://claude.com/claude-code)


arfon:
Verified in production 👍

danielskatz:
thanks!

